PR #1, opened 2015-08-04
Redis sink

ericbottard:


redis-sink/src/main/resources/application.yml
```diff
@@ -0,0 +1,12 @@
+server:
+  port: 8081
+spring:
+  cloud:
+    stream:
+      bindings:
+        input: testtock
```
ericbottard, line 7: I think we should standardize all OOTB modules to have some vanilla string here, something like `demostream`

markfisher, line 7: modules in this repo are "the real thing" so -1 for anything based on the word "demo"

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/CustomRedisConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Creates a dedicated RedisConnectionFactory (that is, different from the spring-cloud-stream one).
+ *
+ * @author Eric Bottard
+ */
+@Configuration
+@ConditionalOnClass({ JedisConnection.class, RedisOperations.class, Jedis.class })
+@EnableConfigurationProperties
+public class CustomRedisConfiguration extends RedisAutoConfiguration {
```
markfisher, line 29: should all of this logic be pushed down to boot itself so that we do not need a custom configuration class?

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/RedisSink.java
```diff
@@ -0,0 +1,102 @@
...
+
+	@Autowired
+	@RedisSinkQualifier
+	private RedisConnectionFactory redisConnectionFactory;
+
+	@Autowired
+	private RedisSinkModuleOptions moduleOptions;
+
+	@Bean
+	public RedisSinkModuleOptions moduleOptions() {
```
markfisher, line 60: why not just use @EnableConfigurationProperties on this class rather than defining a bean?

redis-sink/src/main/resources/application.yml
ericbottard, line 7: Agreed, but one would only ever use this value without overriding it when they'd just run something like "A | B", in a dilettante fashion, no?
As soon as they want to create their real stream(s), they'd need to provide another value.

markfisher, line 7: we need to avoid falling into the trap of thinking that spring-cloud-stream apps will only be used in spring-cloud-data _streams_... instead they should be considered valid as apps in their own right

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/RedisSink.java
ericbottard, line 60: That's supposed to be the same thing (I'm actually realizing I did not put an empty `@EnableConfigurationProperties` on this class). Would not work when not running standalone. Will fix

redis-sink/src/main/resources/application.yml
mbogoevici, line 7: The trouble with using a single value across the board is that it works very well with sources and sinks, but falls apart with processors (which can't use the same value for input and output).

markpollack, line 7: The time and log sink modules currently have a root 'demo' package - so we should change this across the board.  I suggest we use the same package name as where the code lives and just call it <name of module>Application.  for example  org.springframework.cloud.stream.module.redis.RedisSinkApplication.  Package names in time/log are not good either - 'source' and 'sink'.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/CustomRedisConfiguration.java
markpollack, line 29: If we are creating a custom section for the redis properties, then I think this is the way.  I would rename the class RedisSinkAutoConfiguration.  I am not sure that just the 'module' prefix is ok in terms of structure - seems odd.  The ConditionalOnClass should probably include something from this sink, e.g a ref to RedisSink to be a good citizen.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/RedisSink.java
```diff
@@ -0,0 +1,102 @@
...
+	public MessageHandler topicExpressionMessageHandler() {
+		RedisPublishingMessageHandler redisPublishingMessageHandler = new RedisPublishingMessageHandler(redisConnectionFactory);
+		redisPublishingMessageHandler.setTopicExpression(moduleOptions.getTopicExpression());
+		return redisPublishingMessageHandler;
+	}
+
+	@Bean
+	@ConditionalOnExpression("'${module.queue}'.substring(0,1)!='$' || '${module.queueExpression}'.substring(0,1)!='$'")
+	@RedisSinkQualifier
+	public MessageHandler queueExpressionMessageHandler() {
```
markpollack, line 80: The expression in here and the use of the qualifier in so many places feels not very natural.  Maybe we can do something with profiles...

markpollack, line 80: I get that it is to figure out which to create, a topic or a queue, maybe better handled in code.  More concerned with the qualifier atm.

redis-sink/src/main/resources/application.yml
markfisher, line 7: I have changed both `time-source` and `log-sink`, e.g.: https://github.com/spring-cloud/spring-cloud-stream-modules/commit/67bea7588a6802e2658ef86db65e9a457c11fdd9

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/RedisSink.java
ericbottard, line 80: The qualifier is to avoid any "bus" infrastructure code being injected in `messageHandler`, which is a very broad type.

We could duplicate (triplicate actually) the `consumerEndpointFactoryBean()` code and explicitly inject (reference by method call) the actual handler we're interested in.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/CustomRedisConfiguration.java
ericbottard, line 29: The trick is, this class is not an "auto" configuration.

markpollack, line 29: It is an Autoconfiguration, I don't understand.

ericbottard, line 29: It isn't registered as such in spring.factories.

markpollack, line 29: But it should be - it only works now because there is a component scan on that package in the main application class, but if we wanted to reuse this module in another module, it would have to know to component scan that package...  I'm going to take a shot at rewriting this with some changes and post it to my own branch.

ericbottard, line 29: Then maybe this concern should be brought to boot itself (cc @dsyer):
How do I easily create two instances of a whole autoconfiguration, with the second (the non-default one, so to speak) using
- another prefix for `@ConfigurationProperties`
- some way for me to make sure I inject the v2 created beans

I understand this sounds at odds with the concept of AutoConfiguration, but the problem will come up over and over in s-c-s if both the runtime and modules are based off boot

markpollack, line 29: I don't think there is any issue here.  Using another prefix is handled by adding @ConfigurationProperties at the bean level.  The only thing that needs to be distinguished ATM is the redis connection factory used for the sink since the standard namespace 'spring.redis' is being used for the binding.  I think actually making the binding more isolated with a different namespace would be a better way so that frees up them module to be more easily developed - ATM this would effect redis and rabbit, but eventually include kafka and probably the datasource for the batch db.

ericbottard:
Closing, working with https://github.com/spring-cloud/spring-cloud-stream-modules/pull/7 instead

PR #2, opened 2015-08-04
Provide a maven archetype

ericbottard:


markfisher:
This is building in conventions based on what exists in the repo today, and those were copied directly over from the sample modules. They do not represent recommended conventions. In fact, we don't even know what those recommended conventions are yet. Definitely we should not be using "demo" anywhere, and we should not call the application something like "SinkApplication" - again, that is simply a remnant of the "log-sink" being a copy of what was originally the "sink" sample application in spring-cloud-stream.

Overall, I think it's too early to create an archetype.

ericbottard:
Agreed and understood. Yet, I've been building 2 modules in this repo, and copy/pasting is a pain.
The sooner we find answers to "what should we call the application" and what should we use instead of "demo", the better.

What results from applying that archetype can always be changed, so can the archetype itself.

markfisher:
I'd prefer we keep it very generic, with properties like: `basePackage` and `moduleName` (or maybe the artifactId could even be used for the moduleName).

Also I think you would be able to avoid the bug and need for the XXX which then has to be replaced if you are not concatenating.

In general we should not force the concept of source, sink, or processor if this is going to be a general-purpose archetype. However, an _optional_ `moduleType` property could be used for just the class provided to the `@EnableModule` annotation (which is likely to change soon).

I'd also avoid adding things like "blah blah blah" in the documentation. Blank is better IMO.

I would also replace "Options" with "Properties".

markpollack:
Closing this as we should focus on using Spring Initializr instead.  We can bring folks from that team into the discussion to fill any gaps on the Initializr side.

PR #3, opened 2015-08-04
Adding firehose support

viniciusccarvalho:
Adds a doppler firehose source to the modules.

Needed to fix the hierarchy of the poms. spring-cloud-build has an activation profile that makes the maven-exec plugin fail if there's com.google imports on the classpath. 

Discussed this approach with Marius this morning.

I'm still trying to figure out a good way of creating a meaningful integration test, since this requires a websocket server to consume messages from.


markpollack:
Happened to bump into Rossen.  He said to take a look here for some ideas.

https://github.com/rstoyanchev/spring-websocket-portfolio/tree/master/src/test/java/org/springframework/samples/portfolio/web

Note that the WebSocketTestServer implementations were copied out a while ago from test code in Spring Framework itself, so may want to look there to get the latest versions.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "firehose-source/src/main/proto"]
```
mbogoevici: I wish we weren't using submodules. Is there no other alternative?

pom.xml
```diff
@@ -1,51 +1,147 @@
...
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-stream-binder-local</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-stream-binder-redis</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
```
mbogoevici: We shouldn't include both the Redis and Rabbit binder implementations: with both on the CP, there will be a conflict (we need to solve this in spring-cloud-stream)

.gitmodules
viniciusccarvalho: We could just copy the proto files, but I think its more like a pattern I'm seeing on projects that need some sort of external IDL. But I'm ok on removing it if you prefer

firehose-source/README.md
```diff
@@ -0,0 +1,43 @@
+Spring Cloud Stream Firehose Sample
```
markfisher, line 1: we should avoid the term "sample" throughout this file - this is simply a source module

firehose-source/README.md
```diff
@@ -0,0 +1,43 @@
...
+To run this sample, you will need to have installed:
+
+* Java 7 or Above
+* A Redis server
+* [Cloudfoundry](https://github.com/cloudfoundry/cf-release), [Lattice](http://lattice.cf) or a doppler [simulator](https://github.com/viniciusccarvalho/doppler-simulator) running
+* A [protobuf](https://developers.google.com/protocol-buffers) compiler installed and available on your path
+
+## Code Tour
+
+This sample connects to a doppler remote websocket endpoint and emits events from it. The sample converts the binary protocol buffer message into either a XD Tuple or a JSON String, depending on your configuration
```
markfisher: avoid "XD" here... the Tuple class will likely move to a lower level project (e.g. spring-integration) or a "commons" repo (e.g. data-commons)

firehose-source/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    <parent>
+        <artifactId>spring-cloud-stream-modules</artifactId>
+        <groupId>org.springframework.cloud.stream.module</groupId>
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+    </parent>
+
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <start-class>demo.FirehoseApplication</start-class>
```
markfisher: shouldn't use "demo" as the package, rather: `org.springframework.cloud.stream.module.firehose`

firehose-source/src/main/java/source/FirehoseOptionsMetadata.java
```diff
@@ -0,0 +1,119 @@
...
+
+package source;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties(prefix = "firehose")
+public class FirehoseOptionsMetadata {
```
markfisher: should rename: `FirehoseProperties`

firehose-source/src/main/java/source/FirehoseSource.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+@EnableModule(Source.class)
+@Configuration
+public class FirehoseSource implements ApplicationListener<ContextRefreshedEvent>{
+
+    @Autowired
+    private FirehoseOptionsMetadata metadata;
+
+    @Bean
+    public WebSocketInboundChannelAdapter webSocketInboundChannelAdapter(ClientWebSocketContainer webSocketContainer, ByteBufferMessageConverter converter, MessageChannel output) {
```
markfisher: in other cases, we've been autowiring the `Source`, then using source.output() for example to access the channel

firehose-source/src/main/java/source/FirehoseSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @author Vinicius Carvalho
+ */
+@EnableModule(Source.class)
+@Configuration
+public class FirehoseSource implements ApplicationListener<ContextRefreshedEvent>{
+
+    @Autowired
+    private FirehoseOptionsMetadata metadata;
```
markfisher: would rename this field 'properties' (along with the classname change I mentioned above for `FirehoseOptionsMetadata` -> `FirehoseProperties`)

firehose-source/src/main/java/demo/FirehoseApplication.java
```diff
@@ -0,0 +1,109 @@
+/*
```
markfisher: looks like there is an extra line at the beginning of each copyright header

markfisher:
shouldn't need to change the top-level POM, and also should be able to simplify the firehose-source POM now that this has been merged: https://github.com/spring-cloud/spring-cloud-stream-modules/commit/3c8c78eb0cd70e2b9cdb758ef29a403dd5860600

firehose-source/pom.xml
viniciusccarvalho: @markfisher I've updated the project pom, still the imports profile is kicking in, and avoiding the protoc task to run. That was the rationale behind not using spring-cloud-stream as a top project, but instead a dependency. 

Any ideas on how we can solve this issue?

firehose-source/README.md
ericbottard, line 1: I think we should standardize on using asciidoc rather than markdown for those READMEs, as this would ease potential inclusion in our doc toolchain. Also, has better support for variables (which would come handy for _e.g._ replacing versions, etc)

firehose-source/src/main/java/source/FirehoseProperties.java
```diff
@@ -0,0 +1,118 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties(prefix = "firehose")
+public class FirehoseProperties {
+
+    private String dopplerUrl;
```
ericbottard: All of those fields should get Javadoc'd so that metadata can be generated.

firehose-source/src/main/java/source/FirehoseProperties.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+package source;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties(prefix = "firehose")
```
ericbottard: Although we may revisit, the current "best-practice" is to not use a prefix for principal module configuration properties (_i.e._ `--dopplerUrl=foo` rather than `--firehose.dopplerUrl=foo`)

firehose-source/src/main/java/source/FirehoseSource.java
```diff
@@ -0,0 +1,60 @@
...
+        adapter.setMessageConverters(Collections.singletonList(converter));
+        adapter.setOutputChannel(output);
+        adapter.setPayloadType(ByteBuffer.class);
+        return adapter;
+    }
+
+    @Override
+    public void onApplicationEvent(ContextRefreshedEvent contextRefreshedEvent) {
+        ApplicationContext ctx = contextRefreshedEvent.getApplicationContext();
+        ctx.getBean(ClientWebSocketContainer.class).start();
```
ericbottard: Why not have those two beans start in a `@PostConstruct` (or use `InitializingBean`)?

firehose-source/src/main/java/source/FirehoseSource.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@EnableModule(Source.class)
+@Configuration
+public class FirehoseSource implements ApplicationListener<ContextRefreshedEvent>{
+
+    @Autowired
+    private FirehoseProperties metadata;
+
+    @Bean
+    public WebSocketInboundChannelAdapter webSocketInboundChannelAdapter(ClientWebSocketContainer webSocketContainer, ByteBufferMessageConverter converter, MessageChannel output) {
```
ericbottard: Passing `output` like this is not very clear IMHO. Would rather inject

```
@Autowired
private Source channels;
```

and reference `channels.output()`

firehose-source/src/main/java/source/FirehoseSource.java
```diff
@@ -0,0 +1,60 @@
...
+        adapter.setOutputChannel(output);
+        adapter.setPayloadType(ByteBuffer.class);
+        return adapter;
+    }
+
+    @Override
+    public void onApplicationEvent(ContextRefreshedEvent contextRefreshedEvent) {
+        ApplicationContext ctx = contextRefreshedEvent.getApplicationContext();
+        ctx.getBean(ClientWebSocketContainer.class).start();
+        ctx.getBean(WebSocketInboundChannelAdapter.class).start();
```
ericbottard: Moreover, this bean is the defined just above, correct? so definitely don't need lookup

firehose-source/src/main/java/source/TupleFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                .put("timestamp",envelope.getTimestamp())
+                .put("origin",envelope.getOrigin())
+                .put("host",envelope.getIp())
+                .put("index",envelope.getIndex())
+                .put("type",envelope.getEventType().toString())
+                ;
+    }
+
+    /**
+     * Needed due dropsonde decision to use uint64 and little endianess
```
ericbottard: Clean javadoc (dot at the end, remove other tags)

firehose-source/src/main/java/source/TupleFactory.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param uuid
+     * @return
+     */
+    public static String parseUUID(UuidFactory.UUID uuid){
+
+        String uuidStr = null;
+
+        try{
```
ericbottard: Funny formatting. If using IDEA, please see https://github.com/spring-projects/spring-framework/blob/f41de12cf62aebf1be9b30be590c12eb2c030853/src/idea/spring-framework.xml

firehose-source/src/test/java/demo/FirehoseApplicationTests.java
```diff
@@ -0,0 +1,58 @@
...
+    private MessageChannel output;
+
+    @Test
+    public void contextLoads() {
+
+    }
+
+    @ServiceActivator(inputChannel = Source.OUTPUT)
+    public void receive(Message message) {
+        System.out.println(message.getPayload());
```
ericbottard: This will need to change.
There is now support for more meaningful tests in `s-c-s`, see _e.g._ https://github.com/spring-cloud/spring-cloud-stream/blob/3dac86a18acde6e5a489e4f07ad40c7ab1645489/spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/ExampleTest.java

pom.xml
ericbottard: @mbogoevici Actually, with the current autoconf, one will be silently ignored, which is no better

ericbottard:
All classes would require at least type-level javadoc IMHO

firehose-source/src/main/java/demo/FirehoseApplication.java
```diff
@@ -0,0 +1,108 @@
...
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@SpringBootApplication
+@EnableConfigurationProperties(FirehoseProperties.class)
+@ComponentScan(basePackageClasses = FirehoseSource.class)
+public class FirehoseApplication {
```
ericbottard: Is all of this logic typical of this module usage? If so, can it move to the module itself?

firehose-source/src/main/java/demo/FirehoseApplication.java
```diff
@@ -0,0 +1,108 @@
...
+import java.net.URL;
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@SpringBootApplication
+@EnableConfigurationProperties(FirehoseProperties.class)
+@ComponentScan(basePackageClasses = FirehoseSource.class)
```
ericbottard: Once this class has changed package, the component scan will typically not be needed

firehose-source/src/main/java/source/ByteBufferMessageConverter.java
```diff
@@ -0,0 +1,73 @@
...
+        byte[] bytes = new byte[buffer.remaining()];
+        buffer.get(bytes, 0, bytes.length);
+        Tuple result = TupleBuilder.tuple().build();
+        try {
+            EventFactory.Envelope envelope = EventFactory.Envelope.parseFrom(bytes);
+            result = TupleFactory.createTuple(envelope);
+        } catch (InvalidProtocolBufferException e) {
+            e.printStackTrace();
+        }
+        return (metadata.isOutputJson()) ? jsonConverter.convert(result) : result;
```
ericbottard: Wondering if we want to support this, or let this be an s-c-s type conversion feature? cc @markfisher

firehose-source/src/main/java/source/ByteBufferMessageConverter.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public Object convertToInternal(Object payload, MessageHeaders messageHeaders) {
+        return payload;
```
ericbottard: This feels like it violates symmetry, is this intended? maybe throw an exception here if this method is not meant to be called?

firehose-source/src/main/java/source/TupleFactory.java
```diff
@@ -0,0 +1,174 @@
...
+            BigInteger high = new BigInteger(String.valueOf(uuid.getHigh()));
+            byte[] lowBytes = reverse(low.toByteArray());
+            byte[] highBytes = reverse(high.toByteArray());
+            byte[] data1 = new byte[lowBytes.length+highBytes.length];
+            System.arraycopy(lowBytes,0,data1,0,lowBytes.length);
+            System.arraycopy(highBytes, 0, data1, lowBytes.length, highBytes.length);
+
+            StringBuilder builder = new StringBuilder(Hex.encodeHexString(data1));
+            uuidStr = builder.insert(8,"-").insert(13,"-").insert(18,"-").insert(23,"-").toString();
+        }catch (Exception ex){
```
ericbottard: Seems dangerous

viniciusccarvalho: The whole endianness is a nightmare. I looked all over to find a really cleaner way to do that, it does not help the fact that dropsonde is using uint64 and only in java 8 we had long support for that.

I'd happily flame this method if we have a better way. Please let me know and I'll change this. I got stuck for half a day on this madness when I created the old XD firehose.

firehose-source/src/main/java/source/FirehoseProperties.java
viniciusccarvalho: @ericbottard thanks for the comprehensive review of this PR man, this is the only point I'll actually make a point in favor of this use. I think that if one app has more than one module (maybe that never happens), but when accessing the /env endpoint for example, I'd like to see some namespace qualification of where some properties come from. Just my thought on the subject, I'm not part of the team, and its up to you guys to decide the best practices, but I really would like to see prefix on configuration for ease management later.

ericbottard: This whole prefix thing is in (somewhat dormant) discussion. A prefix may be required anyway (for other reasons).
You have a point regarding ease of reading, even though when the module runs alongside other modules, we're likely to provide a way to discern them anyway. But (good) point taken.
For now, I'd remove the prefix so that there is uniformity. We'd of course re-add if this is the decision we take later.

firehose-source/src/main/java/source/FirehoseSource.java
viniciusccarvalho: @ericbottard I need some education here. PostConstruct can not have any arguments, and I really thought that calling wsClient() on a post construct method would return a different instance than the one that was set for the container. If not the case I'm changing the code, and it was a good lesson learned today.

ericbottard: Seems like you found the solution yourself. `@Configuration` classes make
it so that a method call returns the same instance. Could also have
injected the bean in the configuration class itself, amounts to the same,
more or less.
See
http://docs.spring.io/spring/docs/4.2.1.BUILD-SNAPSHOT/spring-framework-reference/htmlsingle/#beans-java-further-information-java-config

Rest assured, I myself learnt things about `@Configuration` not later than
yesterday :)

viniciusccarvalho:
@ericbottard I've fixed everything you've commented on. Thanks for the extensive code review. Still one issue on how our parent pom activates a profile (imports) and that excludes the maven protocol buffer plugin. Not sure how to fix that one

firehose-source/src/main/resources/application.yml
```diff
@@ -0,0 +1,6 @@
+server:
```
ericbottard: Please remove this file and have your 2 defaults (doppler-url & output-json) be set via field initialization

ericbottard:
Hi @viniciusccarvalho, could you please rebase this PR on top of latest master, so that it merges nicely please?
Also, I can't seem to be able to build this. Getting:

```
Downloaded: https://repo.maven.apache.org/maven2/org/apache/maven/maven-model/3.0/maven-model-3.0.jar (161 KB at 221.8 KB/sec)
[WARNING] No 'protocExecutable' parameter is configured, using the default: 'protoc'
[INFO] Compiling 6 proto file(s) to /Users/ericbottard/Documents/projects/spring-cloud-stream-modules/firehose-source/target/generated-sources/protobuf/java
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] spring-cloud-stream-modules ........................ SUCCESS [  1.764 s]
[INFO] Script Variable Generator Support .................. SUCCESS [  2.484 s]
[INFO] spring-cloud-stream-modules-test-support ........... SUCCESS [  1.945 s]
[INFO] spring-cloud-stream-modules-common-configuration ... SUCCESS [  2.958 s]
[INFO] spring-cloud-stream-modules-common ................. SUCCESS [  0.000 s]
[INFO] ftp-source ......................................... SUCCESS [  4.687 s]
[INFO] http-source ........................................ SUCCESS [  5.254 s]
[INFO] time-source ........................................ SUCCESS [  8.244 s]
[INFO] Filter Processor ................................... SUCCESS [  6.360 s]
[INFO] Groovy Filter Processor ............................ SUCCESS [  6.003 s]
[INFO] Groovy Transform Processor ......................... SUCCESS [  5.763 s]
[INFO] Transform Processor ................................ SUCCESS [  5.861 s]
[INFO] counter-sink ....................................... SUCCESS [ 10.596 s]
[INFO] firehose-source .................................... FAILURE [  1.842 s]
[INFO] log-sink ........................................... SKIPPED
[INFO] redis-sink ......................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 01:04 min
[INFO] Finished at: 2015-08-26T10:30:44+02:00
[INFO] Final Memory: 108M/1522M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.google.protobuf.tools:maven-protoc-plugin:0.4.2:compile (default) on project firehose-source: An error occurred while invoking protoc. Error while executing process. Cannot run program "protoc": error=2, No such file or directory -> [Help 1]
```

Does this require come special setup? In which case it should not (we want this repo to build easily, everywhere).

Lastly, regarding the profile activation, seems you've changed the parent, which _may_ be an acceptable solution. I'd rather try to keep it consistent though. What about overriding the enablement of the profile at the child (your pom) level?

OR, and this would cure my problem above I guess, what about removing the protobuf generation from the build altogether, and simply document how the generated java files (that we would check in -- while keeping the `.proto` files checkin in as well) were obtained?

sabbyanandan:
@ericbottard: The compilation problem is perhaps not having _protobuf compiler_ in the classpath? Something that loggregator expects as discussed in this issue [here](https://github.com/cloudfoundry/cf-java-client/issues/163).

ericbottard:
That is what I feared indeed. All the more a reason to remove protobuf from the build and commit the generated files in git (with clear instructions on how to regenerate from `.proto` files).
Could you please do that @viniciusccarvalho (as well as rebasing on top of master)?

ericbottard:
Closing in favor of https://github.com/spring-cloud/spring-cloud-stream-modules/pull/28

PR #4, opened 2015-08-05
Add a Counter Sink that writes to a Boot CounterService

ericbottard:


counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
```diff
@@ -0,0 +1,104 @@
...
+public class CounterSink {
+
+	private static Logger logger = LoggerFactory.getLogger(CounterSink.class);
+
+	private CounterService counterService;
+
+	private CounterSinkOptions options;
+
+	// Default value useful for tests
+	private EvaluationContext evaluationContext = new StandardEvaluationContext();
```
markfisher: why is this here if afterPropertiesSet overrides every time?

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	@Autowired
+	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+		this.beanFactory = beanFactory;
+	}
+
+	@PostConstruct
+	public void afterPropertiesSet() throws Exception {
+		if (!this.evaluationContextSet) {
```
markfisher: should be an explicit null check

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkOptions.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+/**
+ * Holds configuration options for the Counter Sink.
+ *
+ * @author Eric Bottard
+ */
+@ConfigurationProperties
+public class CounterSinkOptions {
```
markfisher: `CounterSinkProperties` would be better IMO

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
ericbottard: I copied this from what we had in XD. I guess the rationale is to have a value that can be used in tests OOTB (hence my comment), but have the SI stuff when running in the real world.

counter-sink/src/main/resources/application.yml
```diff
@@ -0,0 +1,7 @@
+server:
+  port: 8081
+spring:
+  cloud:
+    stream:
+      bindings:
+        input: ticktock
```
markfisher: don't use "ticktock" here

ericbottard: Sure. So, what should I use ;) ?

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
```diff
@@ -0,0 +1,104 @@
...
+		this.options = options;
+	}
+
+	@Autowired
+	public void setCounterService(CounterService counterService) {
+		this.counterService = counterService;
+	}
+
+	@ServiceActivator(inputChannel=Sink.INPUT)
+	public void counterSink(Message<?> message) {
```
markfisher: very confusing having the method name match the class name

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
markfisher: if it's for tests it belongs in test code... i.e. just call the setter in the unit test

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkOptions.java
ericbottard: I really think we should stick with "options", both because this is what we had in XD, and because "properties" feels very "static state describing" (me being French is one of my "properties") whereas "options" carries the idea that you influence behavior (eric.turnOff = true is an "option").

As a last resort, I'd settle for `CounterSinkConfigurationProperties`

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
ericbottard: I agree :) https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSink.java#L35

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkOptions.java
markfisher: IMO there's no need to introduce a term given that these are literally ConfigurationProperties in boot, and this is a boot app (not XD)

counter-sink/src/main/resources/application.yml
markfisher: maybe "counter" or at least something not copied from a completely different context ;)

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkApplication.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A main application that can be used to run the Counter Sink as a standalone app.
+ *
+ * @author Eric Bottard
+ */
+@SpringBootApplication
+public class CounterSinkApplication {
+
+	public static void main(String[] args) throws InterruptedException {
```
ilayaperumalg: Is there a reason why this method throws `InterruptedException`?

counter-sink/src/main/resources/application.yml
ericbottard: I really need to understand what this "default" value is for then.

Is it so that integration tests (which this module currently doesn't have) don't fail?
Or is it so that you can easily wire 2 modules together out of the box?
If the latter, using a module-derived name obviously won't work.

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkApplication.java
ericbottard: Absolutely not (copy/paste). Addressed

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
markfisher: That one was in place since we only had 2 modules "time" and "log". I agree we should change it. This is something we need to determine in general - what is the convention or do we even need to supply a name here vs generating a default (e.g. application.name)? We do need to keep in mind that a spring-cloud-stream module should eventually be runnable as a self-contained app with non-spring-cloud-stream app(s) interacting with it from the other side of the broker.

ericbottard: Now *I* am being confused. Are you saying that method name will turn into a _bean name_. In that case, `counterSink` could make sense.

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
```diff
@@ -0,0 +1,100 @@
...
+
+	private CounterService counterService;
+
+	private CounterSinkOptions options;
+
+	private EvaluationContext evaluationContext;
+
+	private BeanFactory beanFactory;
+
+	@Autowired
```
ilayaperumalg, line 56: Do we really need these setter methods? Can't we just use `@Autowired` on the field values?

ericbottard, line 56: I want to be able to set those from tests. So either `package protected` access, or setters. Some people have very strong opinions about `@Autowired` on fields, I'll introduce them to you :)

This class is a `@Configuration` class, so can't use constructor injection.

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
markfisher: Touché, I will change that one.

counter-sink/src/main/resources/application.yml
markfisher: This is something we need to determine in general - what is the convention or do we even need to supply a name here vs generating a default (e.g. application.name)? We do need to keep in mind that a spring-cloud-stream module should eventually be runnable as a self-contained app with non-spring-cloud-stream app(s) interacting with it from the other side of the broker.

We also need to change from "ticktock" in the log-sink module since that should likewise be a general purpose app. That was in place since we only had 2 modules "time" and "log".

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
mbogoevici, line 56: Can't we bootstrap this as a context from the unit tests themselves, thus supporting DI ?

ericbottard, line 56: I wrote some _unit_ tests.

What you talk about feels more like IT (which I stumbled upon because of `@ConfigurationProperties`)

markfisher, line 56: Autowired on setters seems fine to me.

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
```diff
@@ -0,0 +1,100 @@
...
+	public void setOptions(CounterSinkOptions options) {
+		this.options = options;
+	}
+
+	@Autowired
+	public void setCounterService(CounterService counterService) {
```
ilayaperumalg, line 62: By default it is going to pick up `org.springframework.boot.actuate.metrics.writer.DefaultCounterService`.
How would someone inject a specific implementation of `CounterService`?

ericbottard, line 62: Great question, which needs to be addressed.
It is somehow related to "how to I provide a jdbc driver jar" and similar up-in-the-air questions (like build-time module composition as well. One could envision the user adding a dep on a jar that contains a service bean definition)

markfisher, line 62: Agreed, and once dependencies are contributed, the actual bean instantiation could be driven by `ConditionalOn` annotations.

counter-sink/src/main/resources/application.yml
mbogoevici: We may want to default to applicationName.channelName at SCS level (noting that applicationName is empty, by default) and then skip this property entirely in the modules themselves (with the understanding that the properties are overridable, and document that in the README). That, considering that a PR is the worst place for design discussion :)

ericbottard: It takes a PR and dog food eating to force discussion, so I'd slightly disagree with your last remark :)

Now, trying to understand your comment:

> We may want to default to applicationName.channelName at SCS level

you mean in the `Binder` code itself, right ?

>  and then skip this property entirely in the modules themselves

_i.e._ remove this entirely from this module `application.yml`? I tend to come to that same conclusion, I don't see what this serves (apart maybe from preventing IT tests from failing, as providing an `Environment` in those is not easy)

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSink.java
mbogoevici, line 56: @ericbottard Yes, I think that with this type of component and at this level of granularity, IT and unit tests may overlap a little. 

Maybe isolating the counter() method into a separate class (that can be unit tested in isolation) would be helpful. I feel that we need something similar to Spring MVC Test on the long run for supporting this kind of configuration-with-behaviour class. 

Anyway, looks good overall.

counter-sink/src/main/resources/application.yml
mbogoevici: Yeah, I wasn't suggesting that the discussion was meaningless, just want to move it elsewhere (and reference it), as it doesn't prevent the current PR from being merged. :)

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkOptions.java
markpollack: *Properites is the naming convention in Boot for @ConfigurationProperties classes so we should follow that convention.

ericbottard:
Updated PR

markpollack:
I made changes to support redis as a backend and pushed to master in d0e2720662ab33ec7a1d500797a346906a946552

Known open issues
- A module option to specify inmemory or redis (InMemoryMetricRepository or RedisMetricRepository)
- How to work around/adapt to the name conventions that add "counter." prefix in DefaultCounterService

PR #5, opened 2015-08-07
XD-3283: Add Transform & Filter

ericbottard:


filter-processor/src/main/java/org/springframework/cloud/stream/module/filter/FilterProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+	private FilterProcessorProperties properties;
+
+	@Bean
+	@Transformer(inputChannel = Processor.INPUT)
+	public MessageHandler transformer(BeanFactory beanFactory) throws Exception {
+
+		FilterFactoryBean factoryBean = new FilterFactoryBean();
+		factoryBean.setBeanFactory(beanFactory);
+		factoryBean.setOutputChannel(channels.output());
+		properties.configure(factoryBean);
```
markfisher: IMO the logic that is now in the configure method of the properties class should just be here instead.

filter-processor/src/test/java/org/springframework/cloud/stream/module/filter/FilterProcessorApplicationTests.java
```diff
@@ -0,0 +1,65 @@
...
+
+	@Output(Source.OUTPUT)
+	private MessageChannel output;
+
+	@Input(Sink.INPUT)
+	private MessageChannel input;
+
+	@Test
+	@Ignore
+	public void contextLoads() {
```
markfisher: what is the purpose of this test? (looks like we've copy-and-pasted from tests we had in the actual project, same for log-sink?)

transform-processor/README.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+## Building with Maven
+
+Build the module by executing:
+
+```
+$> mvn package
+```
+
+## Running the Sample
```
markfisher: we need to stop using "sample" and "example" and "demo" in these readmes (and anywhere in the module code)

I actually simplified the READMEs for time-source and log-sink, and I think that's a decent format to use (with more detail as necessary):
https://github.com/spring-cloud/spring-cloud-stream-modules/tree/master/time-source
https://github.com/spring-cloud/spring-cloud-stream-modules/tree/master/log-sink

filter-processor/src/main/java/org/springframework/cloud/stream/module/filter/FilterProcessor.java
ericbottard: I knew you'd say that. I put it there because the Properties class is the one knowing about the kind of transform we're using (so better encapsulation)

ericbottard: https://www.youtube.com/watch?v=WkeNd1Ni4Uw

transform-processor/README.adoc
ericbottard: will update

filter-processor/src/test/java/org/springframework/cloud/stream/module/filter/FilterProcessorApplicationTests.java
ericbottard: first of all, the `@Ignore` should not be there.
This is indeed a copy/paste, but could serve a purpose (the `@Output/@Input` are bogus though): it proves that the module loads with default properties.
Currently working on other tests with non-default props (as discussed in the Slack channel with Stéphane)

filter-processor/src/main/java/org/springframework/cloud/stream/module/filter/FilterProcessor.java
markfisher: re "better encapsulation" I disagree... there are 2 classes: one that provides properties and one that creates beans... creating beans belongs in the one that creates beans I'd say

ericbottard: That's not untrue, but then the one that creates beans will know about the details of which mode we're in (_e.g._ the `@AssertTrue` method is currently in the Properties class [because it's how `@ConfigurationProperties` works, and I think this is good], so IMO that same class should do the `if (script) then ... else ...`). Had the class no validation, I'd be agreement with you (`Properties` class being a dumb data holder).

I'm OK to change it though

markfisher: Validation of two mutually exclusive properties at the time they're being bound is the kind of logic that belongs in the Properties class, I agree.

Deciding how to build a bean based on what properties are set, is the responsibility of the Configuration code. It would be different perhaps if you were not exposing the getters for both of those, but given that, I don't view the "encapsulation" argument as being applicable here.

ericbottard:
Rebased and added tests

markfisher:
What if we break out the groovy script versions of these as `groovy-transformer` and `groovy-filter`, maintaining the non-qualified name for the SpEL variants only? That would considerably simplify the configuration by precluding the conditional aspects. Sure, it's a breaking change, but only for those using the script support, and most likely the SpEL usage is significantly more popular (and the "migration guide" would simply mention if you are using script support, then e.g. move from 'transformer' to 'groovy-transformer')

ericbottard:
Will open another PR with split modules

PR #6, opened 2015-08-11
Redis Sink Refactored

markpollack:
Depends on https://github.com/spring-cloud/spring-cloud-stream/pull/68


markpollack:
squashing

PR #7, opened 2015-08-11
Refactoring of initial redis sink

markpollack:
- Depends on spring-cloud/spring-cloud-stream#68
- Use default spring boot prefix for redis connection factory

### DIFF
--- a/redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkApplication.java
+++ b/redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkApplication.java
@@ -1,0 +1,15 @@
+package org.springframework.cloud.stream.module.redis.sink;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ComponentScan;
+
+@SpringBootApplication
+@ComponentScan(basePackageClasses= RedisSink.class)
+public class RedisSinkApplication {
+
+	public static void main(String[] args) throws InterruptedException {
+		SpringApplication.run(RedisSinkApplication.class, args);
+	}
+
+}
--- a/redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
+++ b/redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
@@ -1,0 +1,66 @@
+package org.springframework.cloud.stream.module.redis.sink;
+
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
+import org.springframework.boot.autoconfigure.redis.RedisAutoConfiguration;
+import org.springframework.boot.autoconfigure.redis.RedisProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.connection.jedis.JedisConnection;
+import org.springframework.data.redis.connection.jedis.JedisConnectionFactory;
+import org.springframework.data.redis.core.RedisOperations;
+import org.springframework.integration.redis.outbound.RedisPublishingMessageHandler;
+import org.springframework.integration.redis.outbound.RedisQueueOutboundChannelAdapter;
+import org.springframework.integration.redis.outbound.RedisStoreWritingMessageHandler;
+import org.springframework.messaging.MessageHandler;
+import redis.clients.jedis.Jedis;
+
+import java.net.UnknownHostException;
+
+/**
+ * Creates a dedicated RedisConnectionFactory different from the one spring-boot autoconfigure
+ * section that the bindings use in spring-cloud-stream.
+ *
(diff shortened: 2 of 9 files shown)


redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+@Configuration
+@ConditionalOnClass({ JedisConnection.class, RedisOperations.class, Jedis.class })
+@EnableConfigurationProperties(RedisSinkProperties.class)
+public class RedisSinkAutoConfiguration {
+
+	@Autowired
+	private RedisConnectionFactory redisConnectionFactory;
+
+	@Autowired
+	private RedisSinkProperties moduleOptions;
```
markfisher, line 47: should rename the `moduleOptions` field to `redisSinkProperties`

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkProperties.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	private String getTopic() {
+		return topic;
+	}
+
+
+	// The javabean property name is what will be reported in case of violation. Make it meaningful
+	@AssertTrue(message = "Exactly one of 'queue', 'queueExpression', 'key', 'keyExpression', 'topic' and 'topicExpression' must be set")
+	public boolean isMutuallyExclusive() {
```
markfisher, line 116: something like this would avoid the need for an extra method:

```
String[] props = new String[] {queue, queueExpression, key, keyExpression, topic, topicExpression};
return (props.length - 1) == Collections.frequency(Arrays.asList(props), null);
```

redis-sink/src/main/resources/application.yml
```diff
@@ -0,0 +1,9 @@
+server:
+  port: 0
+spring:
+  cloud:
+    stream:
+      bindings:
+        input: testtock
```
markfisher, line 7: need to change this

markpollack, line 7: Suggestions?  'Output' assuming binding would prefix with application name?  Need to circle back to the logsink as well with whatever we decide.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
markpollack, line 47: Oops - ok

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkProperties.java
ericbottard, line 116: nice!

redis-sink/src/main/resources/application.yml
ericbottard, line 7: I think the consensus was to simply remove these kinds of section, wasn't it?

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkProperties.java
markpollack, line 116: Reminds me that we should have a test that will cover these code paths.

redis-sink/src/main/resources/application.yml
markpollack, line 7: What about we add a default `spring.application.name: redisSink` and the input default is `input: ${app.name}:input.  We should have this baked in s-c-s so we don't need to put it into the yml explicitly.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkProperties.java
markpollack, line 116: Seems like we should default to one of key, queue, list by default so that I can do `java -jar redisSink-exec.jar` and it runs w/o exception.  Thoughts?

markfisher, line 116: +1 my gut instinct is `'topic=${spring.application.name}'`

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The configuration prefix is "spring.cloud.stream.module.redis" and contains the standard
+ * properties to configure a redis connection, host, port, etc as well as the additional properties for
+ * the sink, queue, key, etc.
+ *
+ * @author Eric Bottard
+ * @author Mark Pollack
+ */
+@Configuration
+@ConditionalOnClass({ JedisConnection.class, RedisOperations.class, Jedis.class })
```
markpollack, line 39: We don't need `@ConditionalOnClass` this as the RedisSink has a compile time dependency on these classes.

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkApplication.java
```diff
@@ -0,0 +1,15 @@
+package org.springframework.cloud.stream.module.redis.sink;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ComponentScan;
+
+@SpringBootApplication
+@ComponentScan(basePackageClasses= RedisSink.class)
```
markpollack, line 8: Can remove `@ComponentScan` as the default package that will be used from `@ComponentScan` in `@SpringBootApplication` already contains RedisSink

redis-sink/src/main/java/org/springframework/cloud/stream/module/redis/sink/RedisSinkAutoConfiguration.java
markpollack, line 47: Will get rid of the autoconfiguration as well, not sure how it will play into module composition etc, so prob better to be conservative until we know we really need it.

markpollack:
Address comments and squashed/merged as 3e51e29f09d29f80e7e5d6756a9b96bdd9346a0b

PR #10, opened 2015-08-13
centralize transport deps and remove obsolete distributionManagement

dturanski:
- Remove distributionManagement from parent pom
- Move redis binder dependency to parent pending https://jira.spring.io/browse/XD-3337. This will avoid rework.


counter-sink/pom.xml
```diff
@@ -26,19 +26,14 @@
...
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-binder-redis</artifactId>
-		</dependency>
-		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-redis</artifactId>
+			<artifactId>spring-boot-configuration-processor</artifactId>
+			<optional>true</optional>
```
ericbottard: This one can move to the parent (and stay, even after 3337)

counter-sink/pom.xml
```diff
@@ -26,21 +26,11 @@
 			<artifactId>spring-cloud-stream</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-binder-redis</artifactId>
-		</dependency>
-		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-redis</artifactId>
```
markfisher, line 30: Can we remove the starter-redis dependency here as well?

markfisher:
should `.settings.xml` be removed as part of this PR also?

ericbottard:
Seems like `.settings.xml` is still to be used. https://github.com/spring-cloud/spring-cloud-stream seems to be the only project of our umbrella where @dsyer has done his maven wrapper cuisine, but from what I understand:
`.settings.xml` is still used for releases. Also, seems repo declarations are to be put in a profile?

dsyer:
.settings.xml is needed for deploying snapshots. The repositories can move to a profile (but it doesn't hurt to leave them in). Look at one of the other cloud projects for a template.

counter-sink/pom.xml
dturanski, line 30: I thought we should leave it because it will need this regardless of the transport used (even though it's redundant now)

markfisher, line 30: makes sense

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/3c8c78eb0cd70e2b9cdb758ef29a403dd5860600

PR #12, opened 2015-08-14
Xd 3283 Ftp Source Module

dturanski:
Ftp Source is the most complex example to date. I would like some early feedback on how this can be made as simple as possible before we do more., the way the common configuration is used, use of SI DSL, etc. 
- spring-cloud-stream-modules-common-configuration -> should these (Mixin replacements, etc.) really go into spring-cloud-stream?
- I don't think we need a separate test-support library here either. 
- S-I DSL seems like the easiest way to implement non trivial flows, but a bit clunkier without lambdas. 
- Some refactoring, renaming of Mixins to align better with the common @Configuration beans
- What's the easiest way to inject module properties? Maybe create instance of the ConfigurationProperties classes and call the setters directly?. Also, some boot supported type conversions are not obvious, so seems better to initialize with a Properties object as the module property values would look to the DSL.  

@ilayaperumalg : From the original commit comment:

> > For the Mixins, I like having the mixins at S-C-S-Modules instead of S-C-S so that there is nothing module specific in there.


dturanski:
@ilayaperumalg  - I removed the original commit comment and replaced with the PR description. Just to clarify your comment re. composition of properties (which I omitted from the description): We discussed in PM that the context is properties for composed modules, which is something we need to figure out, but out of scope for this PR.

ftp-source/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>ftp-source</artifactId>
+	<packaging>jar</packaging>
+	<name>ftp-source</name>
+	<description>FTP Source stream module</description>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<start-class>org.springframework.cloud.stream.module.log.LogSinkApplication</start-class>
```
cppwfs: LogSinkApplication for the start-class?

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSource.java
```diff
@@ -0,0 +1,118 @@
...
+ * @author David Turanski
+ */
+@EnableModule(Source.class)
+@EnableConfigurationProperties({FtpSourceConfigurationProperties.class,
+		FileConsumerConfigurationProperties.class, MaxMessagesConfigurationProperties.class})
+@Import({PeriodicTriggerConfiguration.class, FtpSessionFactoryConfiguration.class})
+public class FtpSource {
+
+	@Autowired
+	private FtpSourceConfigurationProperties config;
```
cppwfs: nitpicky (i know)  config could be renamed ftpSourceConfig

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSourceConfigurationProperties.java
```diff
@@ -0,0 +1,110 @@
...
+@ConfigurationProperties
+public class FtpSourceConfigurationProperties {
+
+
+
+	private String remoteDir = "/";
+
+	private boolean deleteRemoteFiles = false;
+
+	private File localDir = new File("/tmp/xd/ftp");
```
ericbottard: May want to use `java.io.tmpDir` here

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSourceConfigurationProperties.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * @author David Turanski
+ */
+@ConfigurationProperties
+public class FtpSourceConfigurationProperties {
+
+
+
+	private String remoteDir = "/";
```
ericbottard: Add javadoc to all those fields, as this is what will end up in the `.json` metadata file.
Ditto for all other `*Properties` classes

pom.xml
```diff
@@ -37,16 +47,49 @@
...
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.integration</groupId>
+				<artifactId>spring-integration-java-dsl</artifactId>
+				<version>1.0.1.RELEASE</version>
+			</dependency>
+			<dependency>
+				<groupId>org.apache.ftpserver</groupId>
+				<artifactId>ftpserver-core</artifactId>
+				<version>1.0.6</version>
```
ericbottard, line 81: Can force scope=test here.

pom.xml
```diff
@@ -37,16 +47,49 @@
...
+			</dependency>
+			<dependency>
+				<groupId>org.apache.sshd</groupId>
+				<artifactId>sshd-core</artifactId>
+				<version>0.10.1</version>
+			</dependency>
```
ericbottard, line 89: scope = test

spring-cloud-stream-modules-common-configuration/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+		<artifactId>spring-cloud-stream-modules</artifactId>
+		<groupId>org.springframework.cloud.stream.module</groupId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-stream-modules-common-configuration</artifactId>
+	<packaging>jar</packaging>
+	<name>spring-cloud-stream-modules-common-configuration</name>
+	<dependencies>
```
ericbottard: Everything in here should be optional=true (or test scoped) IMHO

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSource.java
```diff
@@ -0,0 +1,118 @@
...
+		switch (fileConsumerConfig.getMode()) {
+			case contents:
+				flowBuilder.enrichHeaders(Collections.singletonMap(MessageHeaders.CONTENT_TYPE,
+						"application/octet-stream"))
+						.transform(new FileToByteArrayTransformer());
+				break;
+			case lines:
+				flowBuilder.enrichHeaders(Collections.singletonMap(MessageHeaders.CONTENT_TYPE, "text/plain"))
+						.split(new FileSplitter(true, fileConsumerConfig.getWithMarkers()), null);
+			default:
```
ericbottard: Please be explicit about the 3rd case (ref) and throw an exception in `default`

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/PropertiesInitializer.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Properties;
+
+import org.springframework.context.ApplicationContextInitializer;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.core.env.PropertiesPropertySource;
+
+/**
+ * @author David Turanski
+ */
+public class PropertiesInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
```
ericbottard: I don't understand the value of this compared to `[Web]IntegrationTest("foo=bar")`, especially if it is used at class level.
What would be really neat (and to be contributed to boot/core), would be method level Environment (re)setting.

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/ftp/TestFtpServer.java
```diff
@@ -0,0 +1,253 @@
...
+	public TestFtpServer(final String root) {
+		this.ftpFolder = new TemporaryFolder() {
+
+			@Override
+			public void create() throws IOException {
+				super.create();
+				ftpRootFolder = this.newFolder(root);
+				sourceFtpDirectory = new File(ftpRootFolder, "ftpSource");
+				sourceFtpDirectory.mkdir();
+				File file = new File(sourceFtpDirectory, "ftpSource1.txt");
```
ericbottard: Hmmm, not sure we want to force the creation of test content in this class. At the very least, extract to some one-liner utility method.

ericbottard: Sorry, I misread the location of this file. IMO, this should be made more generic (ie no default content) and moved to the common project, as it may be used to test several modules (eg ftp source _and_ sink)

ericbottard:
Overall, I like how the DSL reads.
If my test PR is not merged soon, would still be nice if you could rebase on top of it (we'll cherry-pick afterwards) and add tests, to see how it feels.

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/ftp/TestFtpServer.java
dturanski: Yeah. This is copied from XD and originally SI. I noticed XD has some alternate Ftp support, https://github.com/spring-projects/spring-xd/blob/master/spring-xd-test-fixtures/src/main/java/org/springframework/xd/test/fixtures/FtpSource.java#L62

dturanski:
Yes, would like to use the test PR

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSource.java
dturanski: This follows the style of similar examples.  Usually with one ConfigurationProperties set, it is called 'config'. Here we have more than one, so `config` denotes the primary one.

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/PropertiesInitializer.java
markpollack: Yes, method level env setting would be nice.  ATM I think we go with `[Web]IntegrationTest("foo=bar")` and live with the proliferation of test classes until some better approach is found.

time-source/src/main/java/org/springframework/cloud/stream/module/time/TimeSource.java
```diff
@@ -18,37 +18,30 @@
...
 
-	@Bean
-	public Trigger trigger() {
-		PeriodicTrigger trigger = new PeriodicTrigger(properties.getFixedDelay(), TimeUnit.valueOf(properties.getTimeUnit()));
-		trigger.setInitialDelay(properties.getInitialDelay());
-		return trigger;
-	}
-
-	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller(trigger = "trigger", maxMessagesPerPoll = "1"))
+	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller(trigger = PeriodicTriggerConfiguration
```
markpollack: Changes to this TimeSource should not be part of this PR.

time-source/src/main/java/org/springframework/cloud/stream/module/time/TimeSourceProperties.java
```diff
@@ -35,21 +35,6 @@
 	 */
 	private String format = "yyyy-MM-dd HH:mm:ss";
```
markpollack, line 37: Changes to the time source should not be part of this PR

ftp-source/src/test/resources/logback.xml
```diff
@@ -0,0 +1,11 @@
...
+	<appender name="stdout" class="ch.qos.logback.core.ConsoleAppender">
+		<encoder>
+			<pattern>%d{ISO8601} %5p %t %c{2}:%L - %m%n</pattern>
+		</encoder>
+	</appender>
+	<logger name="org.springframework.integration" level="DEBUG"/>
+	<root level="WARN">
+		<appender-ref ref="stdout"/>
+	</root>
+</configuration>
```
markpollack: No newline.

time-source/src/main/java/org/springframework/cloud/stream/module/time/TimeSource.java
dturanski: Just refactoring to use /validate common configuration. It seems like more work to back out and issue another PR to do this but if you feel strongly about it...

ftp-source/src/test/resources/logback.xml
dturanski: ?

markpollack:
Aside from removing changes to the TimeSource and using @IntegrationTest vs. properties initializer, I suggest we merge this and create some new JIRAs to address the following points.
- Use the test binding to assert the output channel has the correct message
- Create a s-c-s-m project that willl hold the @ConfigurationProperties equivalent of XD mixins
- Tests for ftp specific FtpSource @ConfigurationPropertites.
- Investigate how to change application properties at the test method level vs. class method level.

ftp-source/src/test/resources/logback.xml
dturanski: fixed

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/PropertiesInitializer.java
dturanski: I replaced with @IntegrationTest for the simple case.  See FtpSourceApplicationTests, which get some property values from ftpServer methods. I'm not sure if there's a better way in this case but open to suggestions.

dturanski:
- Use the test binding to assert the output channel has the correct message - done already
- Create a s-c-s-m project that willl hold the @ConfigurationProperties equivalent of XD mixins - included here
- Add  JIRA for refactor Time Source to use common configuration

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
```diff
@@ -0,0 +1,91 @@
...
+	@Range(min = 0, max = 65535)
+	public int getPort() {
+		return port;
+	}
+
+	public void setPort(int port) {
+		this.port = port;
+	}
+
+	public String getUsername() {
```
markpollack: This needs to be @NotBlank, otherwise you end up with an exception coming from SI (with 'user' in the message that gives you a false impression the property name is 'user'.

spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationPropertiesTests.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * @author David Turanski
+ */
+public class FtpSessionFactoryConfigurationPropertiesTests {
+
```
markpollack: I think there should be a test that tries to run the app without any env-vars and see that we get a useful error message.  Does an `@ConfigurationProperties` class show all the violations or does it fail fast?

ftp-source/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+
+	<artifactId>ftp-source</artifactId>
+	<packaging>jar</packaging>
+	<name>ftp-source</name>
+	<description>FTP Source stream module</description>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<start-class>org.springframework.cloud.stream.module.log.FtpSourceApplication</start-class>
+	</properties>
```
markpollack, line 18: wrong classname

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
markpollack: Even when adding  `@NotBlank` the validation doesn't seem to kick in 

```
java.lang.IllegalArgumentException: 'user' should be a nonempty string
    at org.springframework.util.Assert.hasText(Assert.java:168) ~[spring-core-4.2.1.BUILD-SNAPSHOT.jar!/:4.2.1.BUILD-SNAPSHOT]
    at org.springframework.integration.ftp.session.AbstractFtpSessionFactory.setUsername(AbstractFtpSessionFactory.java:111) ~[spring-integration-ftp-4.2.0.BUILD-SNAPSHOT.jar!/:na]
    at org.springframework.cloud.stream.module.ftp.FtpSessionFactoryConfiguration.ftpSessionFactory(FtpSessionFactoryConfiguration.java:40) ~[spring-cloud-stream-modules-common-configuration-1.0.0.BUILD-SNAPSHOT.jar!/:1.0.0.BUILD-SNAPSHOT]
    at org.springframework.cloud.stream.module.ftp.FtpSessionFactoryConfiguration$$EnhancerBySpringCGLIB$$2b07175a.CGLIB$ftpSessionFactory$0(<generated>) ~[spring-core-4.2.1.BUILD-SNAPSHOT.jar!/:1.0.0.BUILD-SNAPSHOT]
```

Not sure why.  Also, should we put the JSR303 annotation on the field declaration?

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSource.java
```diff
@@ -0,0 +1,117 @@
...
+import org.springframework.integration.dsl.support.Consumer;
+import org.springframework.integration.file.splitter.FileSplitter;
+import org.springframework.integration.file.transformer.FileToByteArrayTransformer;
+import org.springframework.integration.ftp.session.DefaultFtpSessionFactory;
+import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.scheduling.Trigger;
+
+/**
+ * @author David Turanski
```
markpollack, line 45: A little more javadoc on what the class does...

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
markfisher: Need to add `spring-boot-starter-validation` in the pom.

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.ftp;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.integration.ftp.session.DefaultFtpSessionFactory;
```
markpollack: This dependency is sort of a problem, if I'm building a module that uses 'modules-common-configuration' say for something not related to FTP, I'll get SI+FTP pulled in.  I think we have to separate this into another project/artifact.

markpollack: Ah, I see it is optional - though this led to trouble in the past with spring-batch having optional deps on gemfire, mongo etc....  What do you think?

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/PropertiesInitializer.java
markpollack: I don't see any class with @IntegrationTest on it...   Is there a reason @IntegrationTest can't be used in FtpSourceIntegrationTests?   That seems to be the only class that uses the FTP properties initializer.

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfiguration.java
ericbottard: Well, they weren't really optional, were they? This should be fine, I think (and module consumers need to re-state explicitly what they want).
The alternative is really no better IMO, with a huge proliferation of 1-2 classes artifacts

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
ericbottard: Yeah, there's something weird here. I don't understand how Mark could add `@NotBlank` (so that annotation somehow _was_ on the classpath for some reason) but hibernate-validator was not there. Note that `@NotBlank` is in hibernate-validator, _not_ just in the set of basic jsr303 annotations (which could explain, and would need to be fixed too)

ftp-source/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>ftp-source</artifactId>
+	<packaging>jar</packaging>
+	<name>ftp-source</name>
+	<description>FTP Source stream module</description>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ericbottard, line 16: Can be removed

ftp-source/pom.xml
ericbottard, line 18: I've been bitten wy this as well (somehow IDEA changing more text than it should).
But, why are we even specifying this? If I reckon correctly, the boot plugin can find the main class itself, can't it? Assuming it fails on ambiguity, this may be the best thing to do, wdyt?

spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/PropertiesInitializer.java
dturanski: See my above comment. Maybe static methods on TestFtpServer and use SpEL expressions in the annotations could work, but seems a bit contrived.

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
dturanski: hibernate-validator is on the class path already

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfiguration.java
ericbottard: > Well, they weren't really optional, were they

Or was it because we tried to consume them from gradle?

spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationPropertiesTests.java
dturanski: fail fast

spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryConfigurationProperties.java
```diff
@@ -0,0 +1,92 @@
...
+import org.hibernate.validator.constraints.NotBlank;
+import org.hibernate.validator.constraints.Range;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author David Turanski
+ */
+@ConfigurationProperties
+public class FtpSessionFactoryConfigurationProperties {
```
dturanski: Comparing to https://github.com/spring-cloud/spring-cloud-stream-modules/pull/19 and the boot way, I should remove 'Configuration' from the names or these classes

dturanski:
squashed rebased merged and fixed poms as f4dda5f

PR #13, opened 2015-08-17
XD-3362: Http source

ericbottard:


ericbottard:
Pending merge of https://github.com/spring-cloud/spring-cloud-stream/pull/72 for tests.

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceProperties.java
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * A Map of SpEL expressions to evaluate in order to generate the Message headers.
+	 */
+	private Map<String, Expression> headerExpressions;
+
+	public Map<String, Expression> getHeaderExpressions() {
+		return headerExpressions;
+	}
+
+	public void setHeaderExpressions(Map<String, String> headerExpressions) {
```
ericbottard, line 50: I think we should register a converter for Expressions by default

http-source/README.adoc
```diff
@@ -0,0 +1,31 @@
...
+This module converts incoming http requests into messages.
+
+== Requirements
+
+* Java 7 or Above
+
+== Code Tour
+
+This project is a Spring Boot application that uses Spring Cloud Stream to
+blah blah blah.
```
cppwfs: Insightful :-D

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Bean
+	public Controller inboundController() {
+		HttpRequestHandlingController controller = new HttpRequestHandlingController(false);
+		controller.setRequestChannel(channels.output());
+		controller.setRequestMapping(requestMapping);
+		controller.setPayloadExpression(properties.getPayloadExpression());
+		controller.setHeaderExpressions(properties.getHeaderExpressions());
+		String viewExpression = "new " + DefaultResultView.class.getName() + "()";
+		controller.setViewExpression(new SpelExpressionParser().parseExpression(viewExpression));
```
markfisher, line 58: IIRC you can bypass anything related to views and just return a status code instead (cc @garyrussell). Ideally we would simply return an empty body and ACCEPTED (202).

ericbottard, line 58: I did not find how to do this, the problem being that the `ModelAndView` returned is non-null (and I can't do anything about that).
At the very least, I'll change to 202, but I'd be curious how to do this simpler

markfisher:
Why not just create a simple boot app with a single MVC endpoint, similar to what Dave did in this early sample? (in fact it could be even simpler, calling the channel.send directly instead of using a Gateway): https://github.com/scratches/spring-bus-demo/blob/master/voter/src/main/java/demo/VoterApplication.java#L48

For one thing we want to emphasize that developers can easily create their own spring-cloud-stream modules that are also simple REST apps thereby effectively acting as "http" source modules, but if they only need the simplest passing of HTTP Requests as Messages they can use one we provide out-of-the-box.

ericbottard:
> Why not just create a simple boot app with a single MVC endpoint

because then you can't customize the path (as it would likely be an annotation attribute)

markfisher:
Should be able to use something like: `@RequestMapping("/${path}")`

ericbottard:
Good point.
Yet, this version easily allows multiple paths and everything.

I get your point that we want to showcase ease of use for specific use-cases, but OOTB modules often come as a swiss army knife (the same way transform & filter are complicated because of script support).

Wouldn't the "simple case" better be showcased in -samples?

markfisher:
-1 for swiss-army knife here... since any boot mvc app can become a `spring-cloud-stream` module with the addition of the `@Enable` annotation, we can keep this one to the very simplest case of directly passing messages via a single configurable path.

ericbottard:
Got it a try. We'd have to hardcode the http method if using an annotation, is that ok?

ericbottard:
Opened https://github.com/spring-cloud/spring-cloud-stream-modules/pull/17 as an alternative

PR #15, opened 2015-08-18
Stream modules enhancements

ilayaperumalg:
- Move few more common dependencies to parent POM
  - Move `spring-cloud-stream` and `spring-boot-starter-test` into parent pom.xml.
- Add `remote.name` as a property in application.yml for `spring.cloud.stream.bindings.input/output`
  - This will let the modules connecting to each other using a common `remote.name`
- Misc cleanup

### DIFF
--- a/counter-sink/src/main/resources/application.yml
+++ b/counter-sink/src/main/resources/application.yml
@@ -5,4 +5,4 @@ spring:
   cloud:
     stream:
       bindings:
-        input: counter
+        input: ${remote.name:application:${server.port}}
--- a/counter-sink/pom.xml
+++ b/counter-sink/pom.xml
@@ -15,24 +15,6 @@
 	</parent>
 
 	<properties>
-		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<start-class>org.springframework.cloud.stream.module.metrics.CounterSinkApplication</start-class>
-		<java.version>1.7</java.version>
 	</properties>
-
-	<dependencies>
-		<dependency>
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream</artifactId>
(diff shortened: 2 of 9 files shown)


counter-sink/src/main/resources/application.yml
```diff
@@ -5,4 +5,4 @@ spring:
   cloud:
     stream:
       bindings:
-        input: counter
+        input: ${remote.name:application:${server.port}}
```
ericbottard, line 8: I don't understand this (I'm not even sure how it is parsed, with those two colons).
What is `server.port` doing here? is it a way to provide unicity? I thought we agreed `server.port` would be 0 in this file, to be overridden by s-c-data. Is this the intent here, _i.e._ when `remote.name` is not provided, end up being `application:1234` ?

markfisher, line 8: Actually, `server.port` should be left at its default value (8080) unless the developer of a module has another reason to change that.

markpollack, line 8: I think the standard boot property app.name (afaik) would be a good default

markfisher:
merged but reverted binding names for now via https://github.com/spring-cloud/spring-cloud-stream-modules/commit/223c37676488cea5f6ec7401035f7ee49978bc04

counter-sink/src/main/resources/application.yml
ilayaperumalg, line 8: The idea to introduce `application:${server.port}` as the default value for the binding name was to use the same context Id spring boot [sets](https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/context/ContextIdApplicationContextInitializer.java#L60). Since we use embedded servlet container via spring boot web starter, the `server.port` will always be set.

PR #16, opened 2015-08-18
Add maven wrapper & travis support.

ericbottard:
As discussed with @dsyer, added mvnw support via copy/paste.
Regarding travis, I removed stuff related to docs (which this project doesn't have)
Not sure about the pull-request dichotomy though (how do we expect to publish
those modules? travis or bamboo?)
Also, not sure I understand the rationale regarding `redirectTestOutputToFile`
I'd rather have it to true by default, and false in travis.
cc @spencergibb


ericbottard:
Also, someone with admin access to this repo will need to check the box at travis-ci.org.

markpollack:
merged as 1f98fce7f65ee85bf0b03801c5374b5c56cbf9a6  

Will need to verify travis support for non pr case as using bamboo for snapshot/release build process.

PR #17, opened 2015-08-18
XD-3362: Http source

ericbottard:


http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
```diff
@@ -0,0 +1,50 @@
...
+ * @author Eric Bottard
+ */
+@Controller
+@EnableModule(Source.class)
+public class HttpSource {
+
+	@Autowired
+	private Source channels;
+
+	@RequestMapping(path = "${pathPattern}", method = POST)
```
ericbottard, line 44: Found out that I can't use `path` as a placeholder in `@WebIntegrationTest`, the value of my `$PATH` ENV VAR takes precedence :(

ericbottard, line 44: Also, if we want to provide a default value, will likely need to mess with the Environment inside this configuration.
Moreover, will need to document this with hand written json metadata

http-source/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+	<description>Http Source stream module</description>
+
+	<parent>
+		<groupId>org.springframework.cloud.stream.module</groupId>
+		<artifactId>spring-cloud-stream-modules</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ilayaperumalg: The properties `project.build.sourceEncoding` and `java.version` can be removed as they are taken care at parent POM.

http-source/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+		<start-class>org.springframework.cloud.stream.module.http.HttpSourceApplication</start-class>
+		<java.version>1.7</java.version>
+	</properties>
+
+	<dependencies>
+		<dependency>
```
ilayaperumalg: The dependencies `spring-cloud-stream`, `spring-boot-configuration-processor` and `spring-boot-starter-test` are defined in parent POM as well. I think we can move `spring-cloud-stream-test-support` into parent POM as well after all it is in `test` scope.

http-source/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-http</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-configuration-processor</artifactId>
+			<scope>provided</scope>
```
ilayaperumalg: Curious why `provided` scope here?

http-source/pom.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>http-source</artifactId>
+	<packaging>jar</packaging>
+	<name>Http Source</name>
```
ilayaperumalg, line 7: We need to avoid specifying name with `space` as this name will be used as the binding name.

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
```diff
@@ -0,0 +1,50 @@
...
+	@ResponseStatus(HttpStatus.ACCEPTED)
+	public void handleRequest(@RequestBody byte[] body) {
+		channels.output().send(new GenericMessage<Object>(body));
```
ilayaperumalg, line 47: This may not be the scope of this PR but where would the `messageConverter` be set? Until we provide `messageConverter` as an option, can we use `String` as default so that ContentType would be `text/plain` instead of `application/octet-stream` ? Something like this:

```
channels.output().send(new GenericMessage<String>(new String(body)));
```

http-source/pom.xml
ericbottard, line 7: Wow, did not realize `${name}` in application.yml was actually maven substitution (was amazed that boot capitalized stuff, but did not look further). Thought `name` was contributed to the env by boot itself.
We _really_ need to change the token to _e.g._ `@` make this more obvious.

Will remove in the meantime

ericbottard, line 7: Actually, using ${artifactId} would be better and would address both the format and obviousness

http-source/pom.xml
ericbottard: Pulled up and explained

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
ericbottard, line 47: We need to discuss type conversion at a more general level then. I don't think it makes sense to force String payload.

http-source/src/main/resources/application.yml
```diff
@@ -0,0 +1,8 @@
+server:
+  port: 0
+name: ${spring.application.name:http-source}
```
ericbottard, line 3: Revisiting @ilayaperumalg 's comment about the pom having a space here. Now I know I'm not 100% crazy. Having this file declare a `name` key and also having maven tokenization kick in is _REALLY_ misleading IMO.
I'll submit a PR to change maven's to `@`

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
markfisher, line 44: why? adding `pathPattern` in application.yml works fine

ericbottard, line 44: We need to be careful at how we handle composition/re-use then, as multiple `application.yml` in the same classloader are not taken into account / merged

http-source/src/main/resources/application.yml
```diff
@@ -0,0 +1,8 @@
+server:
```
markfisher, line 1: should be no server.port here (rely on boot default: 8080)

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
mbogoevici, line 44: Wouldn't using a `@PropertySource` with default values work, too?

markfisher, line 44: This works:

```
@PropertySource("classpath:http-source.properties")
public class HttpSourceApplication {
```

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
markfisher, line 47: Default Converters are available, so we could do something to support more types. I'm sure there is a more concise option (cc @rstoyanchev), but for example to support both "text/*" and "application/octet-stream", this works:

```
@RequestMapping(path = "${pathPattern}", method = POST, consumes = "text/*")
@ResponseStatus(HttpStatus.ACCEPTED)
public void handleRequest(@RequestBody String body) {
    channels.output().send(new GenericMessage<String>(body));
}

@RequestMapping(path = "${pathPattern}", method = POST, consumes = "application/octet-stream")
@ResponseStatus(HttpStatus.ACCEPTED)
public void handleRequest(@RequestBody byte[] body) {
    channels.output().send(new GenericMessage<byte[]>(body));
}
```

http-source/src/main/resources/application.yml
markfisher, line 3: we don't want to require module developers to put _anything_ in the yml, so the real solution to the default bindings question needs to happen in the spring-cloud-stream code

http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSource.java
ilayaperumalg, line 47: Along with the above, may be we can return `String` type for `application/xml` and `application/json` along with `text/*`?

ericbottard, line 47: Before we go too far down this road, I think we should discuss type conversion again. In particular for json/xml.

markfisher, line 47: I'll be mapping `application/json` to the String-handling method, but not XML.

markfisher:
Will replace this PR momentarily (with changes discussed above). Testing locally at the moment.

markfisher:
replacing with PR #20 (adds one commit)

PR #21, opened 2015-08-21
Remove application.yml from modules

ilayaperumalg:
- Replace `applocation.yml` with `<module-name>.properties` and import as `@PropertySource` from within module application


ilayaperumalg:
Some of the PropertySource files are in place with `server.port` set to `0` and could be possibly be removed if we assume the developer could set the `server.port` if needed. Also, kept `ticktock` binding name for `time | log` and can be removed as well.

markfisher:
No module should be setting either `server.port` or `spring.cloud.stream.bindings.*` properties. The time-source module does not need a properties file either since the defaults for `fixedDelay` and `initialValue` can be defined in the `TimeSourceProperties`. The same could even be done for the counter-sink so that none of these projects need properties files (only the http-source since it requires the value for ${pathPattern} to be assignable for an annotation attribute).

markfisher:
Just to be clear, what I'm suggesting is that rather than this PR in its current form we should simply delete all of the `application.yml` files for modules and then update `CounterSinkProperties` to handle the default value for name (while taking into account its current support for the mutually exclusive properties: 'name' or 'nameExpression').

ilayaperumalg:
ok, got it. I will get them fixed.

ilayaperumalg:
updated the PR with the suggested changes. thanks!

ericbottard:
> No module should be setting either server.port

I still don't understand what's wrong with setting this to 0. This would allow modules to run out of the box, while still not precluding one (_e.g._ the launcher) from deciding and overriding this. I don't see why writing a module about _e.g._ elastic search sink, I need to worry about having an http connection listening in. 95% of the time, when I run a module (especially as a first-time experience), I don't care about the management features.

markfisher:
The most important principle is "modules are just boot apps", not boot apps _but_ with this one property always set. The default `server.port` for boot apps is 8080 which happens to be the most common default well beyond boot. If you want to make a case for changing boot's default to 0, feel free.

Also, consider container-managed environments (anything from `docker run` to `ltc create` or `cf push`). In those cases, the app is isolated and having a known consistent port is important. That port used in the container is then mapped to a port on the host for a docker container or routed from a URL on the cloud.

ericbottard:
Gotcha (please bookmark this link, so you can point me to it when I ask again in 2 weeks). I'm more convinced by the second point, btw.

markfisher:
Right - the 2nd point is "cloud native" ;) But I'll keep trying 'til I've got you convinced on the "just boot apps" argument as well.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/4428289212411ab7e86b47527dbdf3eda526e29b

PR #22, opened 2015-08-21
XD-3376 Gemfire Sink

dturanski:
Implements a Gemfire Sink. 
- Changed connection properties semantics from XD:  -DhostAddresses=host1:port1,host2:port2 -DconnectType=[locator | server]
- linked with the latest SDG 1.6.2 and Gemfire 8.0.0. Eventually these deps should be added at runtime, same as JDBC modules. We can downgrade the default to 7 perhaps, but upgrade currently requires rebuild per the README.
- Much of this is a port of test fixture code from SDG to launch an external Gemfire Server for IntegrationTests which will be shared by all gemfire modules. I don't think it's worth reviewing here as it's used extensively in the SDG build.


common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfireRegionProperties.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author David Turanski
+ */
+@ConfigurationProperties
+public class GemfireRegionProperties {
+	/**
+	 * The region name.
+	 */
+	@Value("${spring.application.name:}")
```
ericbottard: I'm not sure I understand how the combination of `@ConfigurationProperties` and `@Value` behaves at runtime. Seems to me one has to chose one or the other, or face confusion.

ericbottard:
> Much of this is a port of test fixture code from SDG to launch an external Gemfire Server for IntegrationTests which will be shared by all gemfire modules. I don't think it's worth reviewing here as it's used extensively in the SDG build.

In that case, is there hope for us to maybe re-use an artifact? Maybe SDG could publish their `src/test/java` classes as an artifact?

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfirePoolProperties.java
```diff
@@ -0,0 +1,86 @@
...
+	private boolean subscriptionEnabled;
+	
+
+	@NotEmpty
+	public InetSocketAddress[] getHostAddresses() {
+		return hostAddresses;
+	}
+
+	public void setHostAddresses(String[] hostAddresses) {
+		this.hostAddresses = new InetSocketAddress[hostAddresses.length];
```
ericbottard: May be a bit contrived, but http://stackoverflow.com/a/2347356/1662330 may be the most forward compatible way to "parse"

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfireRegionProperties.java
dturanski: I'm open to suggestion. This is a common pattern in modules, i.e., use stream name as default value for something.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfirePoolProperties.java
dturanski: I saw that but favor the regex.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfireRegionProperties.java
ericbottard: I'm not at all against the _value_ of this annotation and what it achieves. Rather _how_ it achieves it. That is, combining the `@Value` and the `@ConfigurationProperties` approach. Maybe @dsyer has a best practice for those cases where we want a dynamic default value in a `@ConfigurationProperties` bean?

ericbottard: Are there even guarantees about the order in which each mechanism may inject a value?

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfirePoolProperties.java
ericbottard: Maybe we should at least pull that up in a converter (see https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java) so that logic is factored in a single place

dturanski:
> > Maybe SDG could publish their src/test/java classes as an artifact?
- I doubt that this will happen in the next few weeks
- I will refer this to @jxblum (I don't know if he will see this mention automatically). We have considered this before since spring-integration-gemfire has some similar requirements. 
- Note that this code for the most part has little or no dependencies on gem or SDG. It's mostly using ProcessBuilder, etc. to launch an external (JVM) process, wait for it to start and then stopping it. So as a reusable artifact it may rightfully belong elsewhere. However this is a common Gemfire testing story.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfireRegionProperties.java
dturanski: One possibility is to put this in the module (@EnableModule class) itself to use as a default if the property is not set.

jxblum:
Hi Fellas-

Right, the [src/test/java/org.springframework.data.gemfire.process](https://github.com/spring-projects/spring-data-gemfire/tree/master/src/test/java/org/springframework/data/gemfire/process) API is a generalized (GemFire agnostic) way to launch any JVM-based process bootstrapped with a given Java class.  The API stems from ideas/work I did on GemFire (specifically, _Gfsh_), but is even more extensible and reusable.

The (`ProcessExecutor`) API requires a Java class to bootstrap the JVM process.  Currently, I have a few such classes in [src/test/java/org.springframework.data.gemfire.fork](https://github.com/spring-projects/spring-data-gemfire/tree/master/src/test/java/org/springframework/data/gemfire/fork) such as [ServerProcess](https://github.com/spring-projects/spring-data-gemfire/blob/master/src/test/java/org/springframework/data/gemfire/fork/ServerProcess.java), which is used to launch any Spring-configured application (again, GemFire agnostic), given a Spring ApplicationContext configuration file (XML).  An example can be seen [here](https://github.com/spring-projects/spring-data-gemfire/blob/master/src/test/java/org/springframework/data/gemfire/client/ClientCacheSecurityTest.java#L67-L107).

All of this is meant to replace SDG's [ForkUtil](https://github.com/spring-projects/spring-data-gemfire/blob/master/src/test/java/org/springframework/data/gemfire/ForkUtil.java) class as a more capable and general utility for forking JVM processes.

There has been quite a bit of interests (even from customers, e.g. BMO) for having such classes outside of SDG to help them with their (integration) testing efforts.  However, this has been considerably lower priority than other deliverables.

I should have considerably more time after SpringOne to consider this, and I have quite a few more ideas on how to expand and make this API even more robust.

Cheers,
John

mminella:
I'm testing this now with the connected car and came across an issue.  In the connected car we ran the following streams:

```
// Create the initial stream to take HTTP data and enrich it and store it to a local file
stream create --name IoT-HTTP --definition "http | filter --script=file:/opt/pivotal/IoT-ConnectedCar/IoT-DataFilter/src/main/resources/DataFilter.groovy | acmeenrich | shell --inputType=application/json --command='/opt/pivotal/anaconda/bin/python ./StreamPredict.py' --workingDir=/opt/pivotal/IoT-ConnectedCar/IoT-DataScience/PythonModel/ | hdfs" --deploy

// Create a tap that reads the transformed json data into gemfire
stream create --name IoT-Gemfire --definition "tap:stream:IoT-HTTP.shell > typeconversiontransformer | gemfire-server --regionName=car-position --keyExpression=payload.vin" --deploy
```

The `typeconversiontransformer` converts JSON to a java object (`CarPosition` in this case) and returns it.  The gemfire-server sink would then save it to Gemfire.  However, before, we didn't need to add the jar that contains `CarPosition` class to the sink.  In this new version it seems we do.  I'm leaning towards this being a bug with the old version (classpath leak somewhere) but not sure.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfirePoolProperties.java
dturanski: I don't understand this comment

markfisher: I assume @ericbottard is referring to the creation and registration of a `StringToInetSocketAddressConverter` implementation that would encapsulate the regex and be generally reusable.

dturanski:
@mminella - AFAICT with XD you would still need to have the domain object in the sink class path and probably in the Gemfire (the actual server) class path. The recommended approach would be to add that jar to xd/lib. Here that is not an option.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/gemfire/GemfirePoolProperties.java
ericbottard: that was it indeed

dturanski: OK

dturanski: I have a general concern about putting converters like this, or anything that is scoped to a small subset of modules, into s-c-s vs. If we add this to s-c-s-modules-common-configuration, it can easily be implemented as part of this PR. Putting these in s-c-s requires a separate PR on s-c-s for a single class.  That in itself is not a justification, but it is at least a code smell.  I think s-c-s should provide general infrastructure like annotations, serialization, transport, type conversion. This level of module configuration, not so much.

markfisher: ultimately it could go into spring itself. For now, I'd say proceed with the String params for the setter and convert locally

gemfire-sink/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+== Requirements
+
+* Java 7 or above
+* Gemfire 8.0 or above 
+
+Note: This module uses Gemfire v8.0.0. To use with earlier versions of Gemfire, you must change the Gemfire versions in 
+the root pom file and rebuild.
+
+== Code Tour
```
ericbottard: I believe the latest template for section titles is just "Classes", and "Building", "Running" below. cc @markfisher

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/gemfire/GemfireClientRegionConfigurationTests.java
```diff
@@ -0,0 +1,67 @@
...
+	Region<?, ?> region;
+
+	public static class DefaultRegionTests extends GemfireClientRegionConfigurationTests {
+		@Test
+		public void test() {
+			assertThat(region.getName(), equalTo("foo"));
+		}
+	}
+
+	@IntegrationTest({"spring.application.name=foo", "regionName=bar"})
```
ericbottard, line 59: Can remove application name (inherited)

pom.xml
```diff
@@ -160,4 +188,11 @@
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
+	<repositories>
```
ericbottard, line 191: This should go in `settings.xml` I believe

ericbottard:
Did not manage to have this work.
This is what I tried, using the default GemFire server we ship with XD:

```
stream create foo --definition "time | gemfire  --host-addresses=localhost:40404 --region-name=Stocks --key-expression=payload" --deploy
```

I'm getting an "EvaluationContext must not be null" exception. Also, seems like `keyExpression` used to be `<streamname>` in XD, wonder if we can/want replicate this.

dturanski:
what's the stack trace?

ericbottard:
This is a snippet. Not sure if it's actually the root cause, or an exception happening when the error is going to a DLQ or similar:

```
java.lang.IllegalArgumentException: EvaluationContext must not be null
    at org.springframework.util.Assert.notNull(Assert.java:115) ~[spring-core-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
    at org.springframework.expression.spel.ExpressionState.<init>(ExpressionState.java:85) ~[spring-expression-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
    at org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:329) ~[spring-expression-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
    at org.springframework.integration.redis.outbound.RedisQueueOutboundChannelAdapter.handleMessageInternal(RedisQueueOutboundChannelAdapter.java:115) ~[spring-integration-redis-4.2.0.RELEASE.jar!/:na]
    at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:127) ~[spring-integration-core-4.2.0.RELEASE.jar!/:na]
    ... 17 common frames omitted
Wrapped by: org.springframework.messaging.MessageHandlingException: error occurred in message handler [org.springframework.integration.redis.outbound.RedisQueueOutboundChannelAdapter@57b9c7a3]; nested exception is java.lang.IllegalArgumentException: EvaluationContext must not be null
    at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:139) ~[spring-integration-core-4.2.0.RELEASE.jar!/:na]
    at org.springframework.cloud.stream.binder.redis.RedisMessageChannelBinder$1$2.recover(RedisMessageChannelBinder.java:277) ~[spring-cloud-stream-binder-redis-1.0.0.BUILD-SNAPSHOT.jar!/:1.0.0.BUILD-SNAPSHOT]
    at org.springframework.cloud.stream.binder.redis.RedisMessageChannelBinder$1$2.recover(RedisMessageChannelBinder.java:267) ~[spring-cloud-stream-binder-redis-1.0.0.BUILD-SNAPSHOT.jar!/:1.0.0.BUILD-SNAPSHOT]
    at org.springframework.retry.support.RetryTemplate.handleRetryExhausted(RetryTemplate.java:458) ~[spring-retry-1.1.2.RELEASE.jar!/:na]
    at org.sp
```

artembilan:
Hi there! 
Have had a discussion with @dturanski regarding the `EvaluationContext must not be null` error.
Please, find this code:
https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java#L167
and after that this:
https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-spi/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java#L379

So, `onInit()` is called before the setup of `evaluationContext`.

dturanski:
@ericbottard I was able to reproduce this when there is no cache server running. BTW, you need to set `--connect-type=server`. Default is `locator` now which is a breaking change from XD, but more common in the wild.

```
  stream create foo --definition "time | gemfire  --host-addresses=localhost:40404 --connect-type=server --region-name=Stocks --key-expression=payload" --deploy
```

You need a cache server running. For this test, the one the XD distribution one is fine: 

```
      $cd gemfire
      $bin/gemfire-server config/cq-demo.xml
```

The entries will be logged in the console. When you kill the server you (or don't start it, you see `evaluation context cannot be null` which @artembilan tracked to a bug in the way RedisMessageBinder is initialized. (see his comment above)

pom.xml
ericbottard, line 191: Actually, seems like this can be removed altogether, which is better IMO, as we don't want to rely on uncontrolled registries.

trisberg, line 191: Yes, this repo should already be linked to from the Spring IO repository as "gemstone-release-cache"

artembilan:
@dturanski , as we talked with you yesterday: https://jira.spring.io/browse/INT-3850.
I think it will be available in the upcoming `SI-4.2.1` release.

BTW, I don't see reaction to my comment regarding the bug.
Do, you need some ticket on the matter? Where ?
Or do you accept only contribution in face of PRs? 
:smile:

ericbottard:
Merged as b12eaba454a202a91668ee4dd4f7518bebed3616 (as well as https://github.com/spring-cloud/spring-cloud-dataflow/commit/392bc95f4791ce4234e1228f98ce98f043fbbd6d on s-c-d)

PR #24, opened 2015-08-22
Add cloud connector dependencies

ilayaperumalg:
- Add `lattice-connector`, `cf-connector` and `spring-service-connector`
  dependencies to stream modules so that these dependencies need not be added
  into `spring-cloud-stream` or `spring-cloud-stream-binder`.
- This will also let the module developer deploy the self contained modules
  into local, lattice and cf environment without admin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,7 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>1.7</java.version>
+		<spring-cloud-lattice.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-lattice.version>
 	</properties>
 
 	<modules>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -99,12 +100,32 @@
...
 	</dependencyManagement>
-	<!--TODO: Move transport related dependencies here pending https://jira.spring.io/browse/XD-3337  -->
 	<dependencies>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream</artifactId>
 		</dependency>
+		<!-- Cloud connector dependencies -->
+		<!-- Lattice connector dependency to create services info from lattice -->
+		<dependency>
```
markfisher, line 110: why are these 3 dependencies "optional"?

ilayaperumalg, line 110: I think they don't necessarily need to be transitive if the module itself is added as a dependency to other project. and, I believe these cloud dependencies would be left with the developer in that case. what do you think?

markfisher, line 110: That probably does make sense for programmatic composition where modules are embedded, and if they are being embedded into a new module, that itself could (at the developer's discretion) use the parent POM. @mbogoevici  wdyt?

mbogoevici, line 110: I think it's pretty reasonable. A module developer would have the following options:
- new module developers (even outside the project) can extend from scsm-master, thus getting the dependency;
- developers that embed a module and add it as a dependency can explicitly include the dependencies;
- developers that embed a module may add it as parent and get the dependency automatically.

markfisher, line 110: sounds good to me; merging

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/1a97439ad4870b2c49765818d96c8ff129e9349b

PR #25, opened 2015-08-26
Add lattice-core dependency

ilayaperumalg:
- This dependency activates `cloud` and `lattice` profiles when running on Lattice.
- This helps avoiding the need to explicitly activating these profiles when running on Lattice.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,6 +105,13 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream</artifactId>
 		</dependency>
+		<!-- Lattice core dependency that activates cloud,lattice profiles when running on Lattice -->
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-lattice-core</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
+			<optional>true</optional>
+		</dependency>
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
Once this PR is merged, this [PR](https://github.com/spring-cloud/spring-cloud-data/pull/65) on `spring-cloud-data` can be merged as well.

pom.xml
```diff
@@ -105,6 +105,13 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream</artifactId>
 		</dependency>
+		<!-- Lattice core dependency that activates cloud,lattice profiles when running on Lattice -->
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-lattice-core</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
```
markfisher, line 112: do we need the version here now that the cloud starters includes the spring-cloud-lattice project?

ilayaperumalg, line 112: We don't have `spring-cloud-starters` specified in `dependencyManagement` for `s-c-s-modules` since we don't need that for any sub projects but only for the parent POM. Hence the explicit version here.

markfisher, line 112: gotcha

PR #28, opened 2015-08-31
Firehose support

viniciusccarvalho:
Due too many rebase conflicts with the old PR at (https://github.com/spring-cloud/spring-cloud-stream-modules/pull/3) I'm submitting a new PR to address the support for a firehose source.


firehose-source/README.md
```diff
@@ -0,0 +1,47 @@
+Spring Cloud Stream Firehose Sample
+=============================
```
ericbottard, line 2: Could you rephrase this README to have it conform to _e.g._ this template: https://github.com/spring-cloud/spring-cloud-stream-modules/blob/c3d654c2a635474a3f160f62bca0460f91864c0c/counter-sink/README.adoc

Avoid the use of "sample", and simplify the instructions to the maximum.
Use of asciidoc vs markdown would be prefered too.

firehose-source/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+
+    <parent>
+        <artifactId>spring-cloud-stream-modules</artifactId>
+        <groupId>org.springframework.cloud.stream.module</groupId>
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+    </parent>
+
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ericbottard: This can be omitted because inherited

firehose-source/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <start-class>org.springframework.cloud.stream.module.firehose.FirehoseApplication</start-class>
+        <java.version>1.8</java.version>
```
ericbottard: Can this build with java7? if so, this can be omitted because inherited

firehose-source/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    </properties>
+
+    <pluginRepositories>
```
ericbottard: Please remove this repo

pom.xml
```diff
@@ -59,6 +60,35 @@
 			</plugin>
 		</plugins>
 	</build>
+	<!--
+We need to enable this profile for now in order to override the imports execution that removes any dependencies from google.
+
+As of now, the imports activation of spring-cloud-build is not working properly. It activates regardless of the inclusion of the .google file on the source folder.
+-->
+	<profiles>
```
ericbottard, line 68: Can this move down to the firehose pom?

firehose-source/src/main/proto/events/envelope.proto
```diff
@@ -0,0 +1,50 @@
+package events;
```
ericbottard: I see you've removed the protobuf build part from the pom. So is versioning those `.proto` files.

Could you add instructions on how to regenerate the `.java` files from the `.proto` files? In a README in `src/main/proto` for example? Or directly in the top README.
Exact command-line options would be welcome, so that we make sure we can rebuild 2 years from now ... :)

pom.xml
viniciusccarvalho, line 68: Nope, it won't work. I tried that. If its on the firehose pom, build will work on the local module, but will fail from the parent pom

firehose-source/src/main/java/org/springframework/cloud/stream/module/firehose/source/ByteBufferMessageConverter.java
```diff
@@ -0,0 +1,66 @@
...
+    public Object fromMessage(Message<?> message, Class<?> targetClass) {
+        ByteBuffer buffer = (ByteBuffer) message.getPayload();
+        byte[] bytes = new byte[buffer.remaining()];
+        buffer.get(bytes, 0, bytes.length);
+        Tuple result = TupleBuilder.tuple().build();
+        try {
+            EventFactory.Envelope envelope = EventFactory.Envelope.parseFrom(bytes);
+            result = TupleFactory.createTuple(envelope);
+        } catch (InvalidProtocolBufferException e) {
+            e.printStackTrace();
```
ericbottard, line 57: This needs to be better propagated / logged

ericbottard:
Closing this in favor of https://github.com/spring-cloud/spring-cloud-stream-modules/pull/33

Let's continue discussion there

PR #33, opened 2015-09-07
Firehose: Some cleanup over the previous PR

ericbottard:
@viniciusccarvalho:

I have reworked the previous PR quite a bit, mostly formatting and fixing compile issues following the renamings.

Still, I can't get to have it work, trying to target PWS logs. I have a running app on CF, for which I can get logs by `cf logs`.
Also, `grep Doppler ~/.cf/config.json` gives me

```
"DopplerEndPoint": "wss://doppler.run.pivotal.io:4443",
```

so I'm trying to create a stream like so:

```
stream create foo --definition "firehose --dopplerUrl=wss://loggregator.run.pivotal.io:4443 --username=xxx --password=yyy | log" --deploy
```

I'm getting the following error, which may actually be 2 distinct errors in one:

```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'firehoseSource' defined in URL [jar:file:/Users/ericbottard/.m2/repository/org/springframework/cloud/stream/module/firehose-source/1.0.0.BUILD-SNAPSHOT/firehose-source-1.0.0.BUILD-SNAPSHOT-exec.jar!/org/springframework/cloud/stream/module/firehose/source/FirehoseSource.class]: Invocation of init method failed; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'webSocketInboundChannelAdapter' defined in class path resource [org/springframework/cloud/stream/module/firehose/source/FirehoseSource.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.integration.websocket.inbound.WebSocketInboundChannelAdapter]: Circular reference involving containing bean 'firehoseSource' - consider declaring the factory method as static for independence from its containing instance. Factory method 'webSocketInboundChannelAdapter' threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'webSocketContainer' defined in class path resource [org/springframework/cloud/stream/module/firehose/source/FirehoseSource.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.integration.websocket.ClientWebSocketContainer]: Circular reference involving containing bean 'firehoseSource' - consider declaring the factory method as static for independence from its containing instance. Factory method 'webSocketContainer' threw exception; nested exception is java.lang.IllegalArgumentException: Malformed URL for authentication endpoint. Check your configuration.
```

Any reference to a working example would be helpful.

Also, the `dopplerEvents` (which could be changed to `String[]` or even `SomeEnum[]`) prop is never used.
Please use https://github.com/spring-projects/spring-framework/blob/master/src/idea/spring-framework.xml as IDEA formatting rules if adding some commits


viniciusccarvalho:
@ericbottard I'm curious about the compilation issues, I had no issues here, what was failing during the compile phase on your end?

What should I do with that file? I tried to load from intellij but it expects a zip, I've searched all over. On eclipse it was so easy to import a formatting template for a project.

You are missing the `authenticationUrl` property. Note that you will also need the `trustSelfCertificates` if you are running against a PCF deployment that uses self signed certs.

 I have a few changes I was going to submit tomorrow, do I need to rebase from this branch now?

ericbottard:
> @ericbottard I'm curious about the compilation issues, I had no issues here, what was failing during the compile phase on your end?

A couple of variables were used in inner classes w/o being final. Are you compiling in java8? Maybe that would explain

> What should I do with that file? I tried to load from intellij but it expects a zip, I've searched all over. On eclipse it was so easy to import a formatting template for a project.

You need to put it in `~/Library/Preferences/IntelliJIdea14/codestyles`, should then appear as a valid option in this preference window:
screen shot 2015-09-07 at 17 12 24

> You are missing the authenticationUrl property. Note that you will also need the trustSelfCertificates if you are running against a PCF deployment that uses self signed certs.

I'll look into this. Any idea what this would be when running against Pivotal Web Services?

> I have a few changes I was going to submit tomorrow, do I need to rebase from this branch now?

Yes please, there have been quite a lot of changes in the upstream project

viniciusccarvalho:
@ericbottard I've never tried to get it working on PWS. AFAIK you need an user with doppler.read permissions. Not sure if the regular user of pws has that.

I've tested with lattice, the simulator and my own PCF env (after adding doppler.read to an user) but never with PWS.

So, my changes were on the README and the exception propagation of the transformer, which seems that you've already fixed. 

What do you prefer? I need to rebase on your repo and merge your branch now? 

I have a couple of hours tomorrow I could ping you to discuss.

ericbottard:
Let's do that yes. Ping me on slack

Best thing would be to issue a PR against my branch for example (having
rebased on top of mine)

markpollack:
after chatting w/ Vinicius seems like we should close this one for now and wait for another one to roll in based off eric's branch.

PR #34, opened 2015-09-09
Add Twitterstream module

ilayaperumalg:
- Supports both `sample` and `firehose` streams
- Tightly aligned with `spring-social-twitter`'s TwitterTemplate's operations implementations

### DIFF
--- a/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterCredentials.java
+++ b/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterCredentials.java
@@ -25,12 +25,24 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 //todo: Move these properties into spring-boot's TwitterProperties prefixed with: `spring.social.twitter`
 public class TwitterCredentials {
 
+	/**
+	 * Consumer key
+	 */
 	private String consumerKey;
 
+	/**
+	 * Consumer secret
+	 */
 	private String consumerSecret;
 
+	/**
+	 * Access token
+	 */
 	private String accessToken;
 
+	/**
+	 * Access token secret
+	 */
 	private String accessTokenSecret;
 
 	public String getConsumerKey() {
--- a/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterCredentials.java
+++ b/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterCredentials.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2015 the original author or authors.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.module.twitter;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Twitter credentials.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@ConfigurationProperties
+//todo: Move these properties into spring-boot's TwitterProperties prefixed with: `spring.social.twitter`
+public class TwitterCredentials {
+
+	private String consumerKey;
+
+	private String consumerSecret;
+
+	private String accessToken;
+
+	private String accessTokenSecret;
+
+	public String getConsumerKey() {
+		return this.consumerKey;
+	}
+
+	public void setConsumerKey(String consumerKey) {
+		this.consumerKey = consumerKey;
+	}
+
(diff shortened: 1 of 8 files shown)


common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterCredentials.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Twitter credentials.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@ConfigurationProperties
+//todo: Move these properties into spring-boot's TwitterProperties prefixed with: `spring.social.twitter`
+public class TwitterCredentials {
+
+	private String consumerKey;
```
ericbottard, line 31: Need to javadoc all the fields so that we generate json metadata

twitterstream-source/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterStreamMessageProducer.java
```diff
@@ -0,0 +1,98 @@
...
+			public void onLimit(int numberOfLimitedTweets) {
+				logger.info("The stream is being track limited for " + numberOfLimitedTweets + " tweets.");
+			}
+
+			@Override
+			public void onWarning(StreamWarningEvent warningEvent) {
+				logger.error("Streaming source is in the danger of being disconnected.");
+			}
+		});
+		if (this.streamType.equalsIgnoreCase(TwitterStreamProperties.FIREHOSE)) {
```
ericbottard: why not use an enum?

twitterstream-source/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterStreamProperties.java
```diff
@@ -0,0 +1,49 @@
...
+public class TwitterStreamProperties {
+
+	//todo: Support more stream types such as `filter`, `search` that TwitterTemplate implementation supports.
+	protected static final String SAMPLE = "sample";
+
+	protected static final String FIREHOSE = "firehose";
+
+	private String streamType = SAMPLE;
+
+	@AssertTrue(message = "Only `sample` or `firehose` stream operations are supported.")
```
ericbottard: see my message regarding the use of an enum.

BUT, if we keep this String style, it's better to turn things around, as the name of the getter (`streamTypeValid`) will be reported, which is not very easy to grasp. It's better to do

```
@AssertFalse(message=...)
boolean isInvalidStreamType() {
}
```

twitterstream-source/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterStreamMessageProducer.java
```diff
@@ -0,0 +1,98 @@
...
+			twitterTemplate.streamingOperations().firehose(listeners);
+		}
+		else {
+			twitterTemplate.streamingOperations().sample(listeners);
+		}
+	}
+
+	@Override
+	public void onInit() {
+		setOutputChannel((MessageChannel) getBeanFactory().getBean(Source.OUTPUT));
```
ericbottard: this feels wrong.
Should be passed in the `@Configuration` IMHO

markfisher: +1 call setOutputChannel in the `@Bean` method that creates this in `TwitterStreamSource` (the channel itself can be autowired into that Source class).

ilayaperumalg: +1

twitterstream-source/src/main/java/org/springframework/cloud/stream/module/twitter/TwitterStreamProperties.java
ilayaperumalg: I think it's better to keep the `streamType` as String rather than enum for better error handling. I will incorporate the change to make the assertion method to read `invalidStreamType`. What do you think?

ericbottard:
> I think it's better to keep the streamType as String rather than enum for better error handling. I will incorporate the change to make the assertion method to read invalidStreamType. What do you think?

I'm not sure I get what you mean "for better error handling". I would advocate for the contrary: using an enum, we get validation for free, as well as hints for the possible values eventually. And this is a case where we're in total control of the enum values, so I see no downside here

ilayaperumalg:
@ericbottard  When an invalid enum value is used, the boot conversion fails this way (for the invalid enum `test`):

```
[Failed to convert property value of type 'java.lang.String' to required type 'org.springframework.cloud.stream.module.twitter.TwitterStreamType' for property 'streamType'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type java.lang.String to type org.springframework.cloud.stream.module.twitter.TwitterStreamType for value 'test'; nested exception is java.lang.IllegalArgumentException: No enum constant org.springframework.cloud.stream.module.twitter.TwitterStreamType.test]
```

I like the error messages when using string rather than using enum.
Do you have any suggestion to make this better?

ericbottard:
Could go as far as submit a PR on spring-core to cite all possible values when dealing with an enum.

But more importantly, I think we should handle this consistently, and wrap it with more user friendly interpretation if we deem it necessary. The caller (us) has all the info to better explain what the problem is. The implementer (the module author) should not need to bother

ericbottard:
Merged as f6f67a9 (alongside https://github.com/spring-cloud/spring-cloud-dataflow/commit/4a855da24114ca45b5b64108912ff48888a74a75).
A `README.adoc`would be welcome, with a tiny paragraph about obtaining keys (separate PR)

PR #35, opened 2015-09-10
XD-3468 Adding missing task scheduler/executor to configuration

trisberg:


hdfs-sink/src/main/java/org/springframework/cloud/stream/module/hdfs/sink/DataStoreWriterFactoryBean.java
```diff
@@ -186,6 +195,16 @@ public void setProperties(HdfsSinkProperties properties) {
 		this.properties = properties;
 	}
 
+	@Autowired
+	public void setScheduler(TaskScheduler scheduler) {
```
jvalkeal: any reason why not use `setTaskScheduler` as method name for this and `setTaskExecutor` for below executor?

PR #36, opened 2015-09-10
Adding hdfs sink to list of available modules

trisberg:


sabbyanandan:
The 'Available Modules' section-format is changed from list to table and the **hdfs** sink has been included. This PR is not applicable anymore.

PR #38, opened 2015-09-10
Twitterstream in line with XD 1.x

ilayaperumalg:
- The changes in this PR will have the twitter stream reader that XD 1.x uses so that the compatibility in mapping the json output is intact
  - Use the custom adapter that reads the twitterstream data directly using the twitter REST template (not using the StreamTemplate)


trisberg:
Ran fine for me and tweets seeem to match format in XD 1.x

PR #41, opened 2015-10-01
add file-sink

markfisher:


file-sink/src/main/java/org/springframework/cloud/stream/module/file/FileSinkProperties.java
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Properties for the file sink.
+ *
+ * @author Mark Fisher
+ */
+@ConfigurationProperties
+public class FileSinkProperties {
+
+	private static final String DEFAULT_DIR =
+			File.separator + "tmp" +
```
ericbottard: I don't think this is sufficient to work on windows, is it?
This will result in `\tmp\dataflow\output`, that is all.

Should somehow do `${java.io.tmpDir} + /dataflow/output` instead

file-sink/src/main/java/org/springframework/cloud/stream/module/file/FileSinkProperties.java
```diff
@@ -0,0 +1,158 @@
...
+			File.separator + "dataflow" +
+			File.separator + "output";
+
+	private static final String DEFAULT_NAME = "file-sink";
+
+	/**
+	 * whether content should be written as bytes
```
ericbottard: Given this is javadoc, the "new" way of writing those will be Capital at start + dot at end (ie the opposite of XD), to be consistent throughout all `@ConfigurationProperties` (whether we choose to alter this a bit at _display time_ is a different story, but I think we won't)

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
```diff
@@ -0,0 +1,93 @@
...
+	@Autowired
+	protected Sink sink;
+
+	@WebIntegrationTest({"name = test", "dir = /tmp/dataflow-tests", "suffix=txt"})
+	public static class TextTests extends FileSinkTests {
+
+		@Test
+		public void test() throws Exception {
+			sink.input().send(MessageBuilder.withPayload("this is a test").build());
+			File file = new File("/tmp/dataflow-tests/test.txt");
```
ericbottard: Won't work on windows

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
```diff
@@ -0,0 +1,93 @@
...
+			file.deleteOnExit();
```
ericbottard, line 61:
```
    @Rule
    public TemporaryFolder folder= new TemporaryFolder();
```

and adding the file in that folder is the more idiomatic way.

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
```diff
@@ -0,0 +1,93 @@
...
+	public static class BinaryTests extends FileSinkTests {
+
+		@Test
+		public void test() throws Exception {
+			sink.input().send(MessageBuilder.withPayload("foo".getBytes()).build());
+			File file = new File("/tmp/dataflow-tests/file-sink");
+			file.deleteOnExit();
+			assertTrue("file does not exist", file.exists());
+			byte[] results = FileCopyUtils.copyToByteArray(file);
+			assertEquals(3, results.length);
```
ericbottard, line 77: Can use `Assert.assertArrayEquals()`

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
```diff
@@ -0,0 +1,93 @@
...
+			assertEquals((byte)'f', results[0]);
+			assertEquals((byte)'o', results[1]);
+			assertEquals((byte)'o', results[2]);
+		}
+	}
+
+	@WebIntegrationTest({"nameExpression = payload.substring(0, 4)",
+			"dirExpression = '/tmp/dataflow-tests/'+headers.dir", "suffix=out"})
```
ericbottard: Won't work on windows, but this is one trickier :( May want to make it conditional on *nix with a `Rule` or something

file-sink/README.md
```diff
@@ -0,0 +1,35 @@
+Spring Cloud Stream File Sink
+=============================
```
ericbottard: Please use asciidoc, pretty please. 

In all cases, it's better to avoid this underlining form for titles, very hard to deal with in an automatic fashion

file-sink/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+This module writes each message it receives to a file.
+
+## Requirements:
+
+* Java 7 or Above
+* Redis running on localhost
```
ericbottard: Do we still want that reference to redis?

file-sink/src/main/java/org/springframework/cloud/stream/module/file/FileSinkProperties.java
markfisher: Okay. I will change it. I only followed this format because I _thought_ it was the "new" way since I used a module that was just added a few days ago as my example :)

file-sink/README.md
markfisher: no - will remove that (in fact I don't think we need this section at all)

file-sink/README.md
markfisher: sure - we should change the others as well I guess

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
markfisher, line 61: I need to have an explicitly named dir in the annotations. Would that be possible if a `TemporaryFolder` is the root of the path?

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
markfisher, line 77: Cool.

file-sink/src/main/java/org/springframework/cloud/stream/module/file/FileSinkProperties.java
ericbottard: Which one? I did not see this.

markfisher: https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/hdfs-sink/src/main/java/org/springframework/cloud/stream/module/hdfs/sink/HdfsSinkProperties.java

trisberg: I was just following the TimeSourceProperties style - maybe we need to update the "Spring Cloud Stream Module Development" guide

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
trisberg, line 61: I did something similar in the HDFS sink tests - https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/hdfs-sink/src/test/java/org/springframework/cloud/stream/module/hdfs/sink/HdfsSinkIntegrationTests.java#L54

Also, should we use `dir` or `directory` for the property? HDFS sink uses `directory`.

markfisher, line 61: I would prefer 'directory', but at least for the initial version I simply ported every property as-is from the XD 1.x file sink.

ericbottard, line 61: I would prefer `directory` as well.
One thing to keep in mind when dealing with these kind of issues is that Boot property metadata uses gnu-style syntax by default (which goes well with our `--` prefix). So full words, and dash separated will become the canonical norm:

```
--dir-expression 
```

or maybe even

```
--directory-expression
```

file-sink/src/main/java/org/springframework/cloud/stream/module/file/FileSinkProperties.java
markfisher: +1

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
markfisher: Fixing all test paths.

file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java
markfisher, line 61: Changing the properties to `directory` and `directoryExpression`.

ilayaperumalg:
LGTM, merged.

ilayaperumalg:
Also updated the `file-sink` to spring-cloud-dataflow registry.

PR #42, opened 2015-10-05
[Sprint: 59] XD-3412: Port SFTP Source from XD

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3412

`s/user/username/` for consistency with the FTP source and avoid conflict with USER env variable.


garyrussell:
Closing for now; reworking with changes in the XD module (`knownHostsExpression`) and also refactoring to make SFTP and FTP consistent.

common/spring-cloud-stream-modules-common-configuration/pom.xml
```diff
@@ -29,6 +29,12 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-sftp</artifactId>
```
mbogoevici, line 43: Why is the SFTP module (and FTP module for that matter) added as part of the common configuration?

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
```diff
@@ -98,6 +99,7 @@ public void noPropertiesThrowsMeaningfulException() {
 			context.register(Conf.class);
 			context.refresh();
 			fail("should throw exception");
+			context.close();
```
mbogoevici: Is this reachable? Seems like in either case an exception is thrown before getting here. Maybe we should close in the `finally` clause.

common/spring-cloud-stream-modules-common-configuration/pom.xml
garyrussell, line 43: I followed the FTP model - they are in common because the connection factory configuration is common across source and sink.

They are marked as optional so they won't be pulled in when not needed.

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
garyrussell: It's just there to stop STS (eclipse) from barking at me about a `Closeable` that's not closed. I just hate yellow squiggly lines :smile: 

It could go before the `fail`, I guess, but then it might imply it's the close the that's expected to fail.

mbogoevici: Ok, let's keep STS happy :)

common/script-variable-generator/.gitignore
```diff
@@ -0,0 +1 @@
+/.apt_generated/
```
ericbottard: Maybe we can pull these up

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+public abstract class AbstractRemoteFileSourceProperties {
+
+	/**
+	 * The remote FTP directory, "/" by default.
+	 */
+	private String remoteDir = "/";
+
+	/**
+	 * Set to true to delete remote files after successful transfer, false by default.
```
ericbottard: defaults should not be documented through javadoc (they will be presented to user by other means)

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,144 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.module.file.remote;
+
+import java.io.File;
+
+/**
```
ericbottard, line 26: Should add javadoc

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * The suffix to use while the transfer is in progress, ".tmp" by default.
+	 */
+	private String tmpFileSuffix = ".tmp";
+
+	/**
+	 * A filter pattern to match the names of files to transfer, no filter by default.
+	 */
+	private String filenamePattern;
```
ericbottard, line 63: Is this mutually exclusive with `filenameRegex`? If so, exclusion should be implemented by jsr303 validation

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,144 @@
...
+package org.springframework.cloud.stream.module.file.remote;
+
+import java.io.File;
+
+/**
+ * @author David Turanski
+ * @author Gary Russell
+ *
+ */
+public abstract class AbstractRemoteFileSourceProperties {
```
ericbottard, line 33: Globally, defaults should not be documented by javadoc.
Also, jsr303 validation rules would be welcome where appropriate

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteServerProperties.java
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.module.file.remote;
+
+import org.hibernate.validator.constraints.NotBlank;
+import org.hibernate.validator.constraints.Range;
+
+/**
+ * @author David Turanski
```
ericbottard, line 23: Javadoc (and maybe even classname / package) should specify in which context this class applies. Default port seems to imply FTP

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.integration.sftp.session.DefaultSftpSessionFactory;
+
+/**
+ * @author Gary Russell
+ */
+@Configuration
+@EnableConfigurationProperties(SftpSessionFactoryProperties.class)
+public class SftpSessionFactoryConfiguration {
+	public static final String SFTP_SESSION_FACTORY_BEAN_NAME = "sftpSessionFactory";
```
ericbottard: This (public) does not seem to be referenced from elsewhere. Is it really necessary?

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryProperties.java
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * True to allow an unknown or changed key.
+	 */
+	private boolean allowUnknownKeys = false;
+
+	/**
+	 * A SpEL expression resolving to the location of the known hosts file.
+	 */
+	private String knownHostsExpression = null;
```
ericbottard, line 52: Would there be any value in making this of type `Expression`? See https://github.com/spring-cloud/spring-cloud-stream-modules/blob/b12eaba454a202a91668ee4dd4f7518bebed3616/transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/TransformProcessor.java#L35
At least would bring early parsing.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryProperties.java
```diff
@@ -0,0 +1,79 @@
...
+package org.springframework.cloud.stream.module.sftp;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cloud.stream.module.file.remote.AbstractRemoteServerProperties;
+
+/**
+ * @author Gary Russell
+ */
+@ConfigurationProperties
+public class SftpSessionFactoryProperties extends AbstractRemoteServerProperties {
```
ericbottard, line 27: Add some jsr303 validation? some of the props have defaults of `""` instead of `null`, which seems to call for `@NotNull` at least

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
```diff
@@ -98,6 +99,7 @@ public void noPropertiesThrowsMeaningfulException() {
 			context.register(Conf.class);
 			context.refresh();
 			fail("should throw exception");
```
ericbottard: Can this changed to `@Test(expected=)` ?

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
ericbottard: Maybe go in in finally block then?

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
ericbottard: From the catch block below, I guess it reads OK as it is now (although `ExpectedException` works too). Please disregard

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryPropertiesTests.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.validation.FieldError;
+
+/**
+ * @author David Turanski
+ * @author Gary Russell
+ */
+public class SftpSessionFactoryPropertiesTests {
+
+	@Test
+	public void hostCanBeCustomized() {
```
ericbottard, line 42: Seems that all those tests are actually testing spring core / boot.

common/spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/file/remote/TestFtpServer.java
```diff
@@ -159,14 +161,17 @@ public void before() throws Throwable {
 
 
 	@PreDestroy
-	public void after() {
-		this.server.stop();
+	public void after() throws Exception {
+		stopServer();
 		this.ftpTemporaryFolder.delete();
 		if (localTemporaryFolder != null) {
 			this.localTemporaryFolder.delete();
```
ericbottard, line 168: This is done for you by `TemporaryFolder` already.
Seems it is missused though, I don't see no `@Rule`

common/spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/file/remote/TestSftpServer.java
```diff
@@ -38,20 +41,73 @@
 
 	private final SshServer server = SshServer.setUpDefaultServer();
 
-	private final int port;
-
 	public TestSftpServer(String root, int port) {
 		super(root);
-		this.port = port;		
+		setFtpTemporaryFolder(
+				new TemporaryFolder() {
```
ericbottard, line 47: Hmmm, yeah, seems like missuse of `TemporaryFolder`.
Also, forces some paths in `TestSftpServer` which makes it less re-useable. Is there a better way?

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSource.java
```diff
@@ -77,15 +82,24 @@ public PollerMetadata poller() {
...
 				.temporaryFileSuffix(config.getTmpFileSuffix())
-				.deleteRemoteFiles(config.isDeleteRemoteFiles())
+				.deleteRemoteFiles(config.isDeleteRemoteFiles());
+
+		if (StringUtils.hasText(this.config.getFilenamePattern())) {
+			messageSourceBuilder.filter(new FtpSimplePatternFileListFilter(this.config.getFilenamePattern()));
+		}
+		else if (StringUtils.hasText(this.config.getFilenameRegex())) {
+			messageSourceBuilder
+					.filter(new FtpRegexPatternFileListFilter(Pattern.compile(this.config.getFilenameRegex())));
```
ericbottard: Should make the property a `Pattern` itself (adding a dedicated Converter if none already exists). This would surface pattern compilation errors early

ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSourceProperties.java
```diff
@@ -15,118 +15,14 @@
...
-	}
-
-	public void setRemoteFileSeparator(String remoteFileSeparator) {
-		this.remoteFileSeparator = remoteFileSeparator;
-	}
-
-	public boolean isPreserveTimestamp() {
-		return preserveTimestamp;
-	}
+public class FtpSourceProperties extends AbstractRemoteFileSourceProperties {
```
ericbottard, line 26: Maybe this is a way to address my earlier comment about port 21. Have dedicated subclasses?

ftp-source/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSourceIntegrationTests.java
```diff
@@ -103,6 +104,7 @@ public void create() throws IOException {
 		properties.put("remoteDir", ftpServer.getSourceFtpDirectory().getName());
 		properties.put("username", "foo");
 		properties.put("password", "foo");
+		properties.put("filenamePattern", "*");
 		properties.put("port", ftpServer.getPort());
 		properties.put("mode", "ref");
 		PropertiesInitializer.PROPERTIES = properties;
```
ericbottard, line 110: This should be replaced with support by `@[Web]IntegrationTest`

sftp-source/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>sftp-source</artifactId>
+	<packaging>jar</packaging>
+	<name>sftp-source</name>
+	<description>SFTP Source stream module</description>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ericbottard: This is inherited

sftp-source/src/main/java/org/springframework/cloud/stream/module/sftp/source/SftpSource.java
```diff
@@ -0,0 +1,133 @@
...
+	MaxMessagesProperties maxMessagesProperties;
+
+	@Autowired
+	DefaultSftpSessionFactory sftpSessionFactory;
+
+	@Autowired
+	Trigger trigger;
+
+	@Autowired
+	@Bindings(SftpSource.class)
```
ericbottard: This is not needed as far as I understand (but does not hurt)

sftp-source/src/main/java/org/springframework/cloud/stream/module/sftp/source/SftpSource.java
```diff
@@ -0,0 +1,133 @@
...
+	Source source;
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.trigger(this.trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
+	}
+
+	@Bean
+	public IntegrationFlow sftpInboundFlow() {
```
ericbottard, line 81: I'm not familiar enough with SI DSL. What makes this context actually work? Are `IntegrationFlow` beans handled in some way? I don't see any SI annotation here.

sftp-source/src/test/java/org/springframework/cloud/stream/module/sftp/source/SftpSourceIntegrationTests.java
```diff
@@ -0,0 +1,111 @@
...
+				+ sftpServer.getSourceFtpDirectory().getName());
+		properties.put("localDir", sftpServer.getRootFolder().getAbsolutePath() + File.separator
+				+ sftpServer.getTargetLocalDirectory().getName());
+		properties.put("username", "foo");
+		properties.put("password", "foo");
+		properties.put("port", sftpServer.getPort());
+		properties.put("mode", "ref");
+		properties.put("allowUnknownKeys", "true");
+		properties.put("filenameRegex", ".*");
+		PropertiesInitializer.PROPERTIES = properties;
```
ericbottard, line 86: Replace with `@IntegrationTest`

ericbottard, line 86: Except you can't because of `sftpServer.getRootFolder().getAbsolutePath() + File.separator+ sftpServer.getTargetLocalDirectory().getName()`  :(

sftp-source/src/test/resources/logback.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
```
ericbottard: Is this really needed?

sftp-source/src/main/java/org/springframework/cloud/stream/module/sftp/source/SftpSource.java
artembilan, line 81: There is just enough to have somewhere in your ctx `@EnableIntegration` and `IntegrationConfigurationBeanFactoryPostProcessor` will load all `IntegrationConfigurationInitializer` impls, including `DslIntegrationConfigurationInitializer` through the `spring.factory` mechanism. 
And the last one adds `IntegrationFlowBeanPostProcessor` to post-process `IntegrationFlow` beans.

ericbottard, line 81: Thx @artembilan , makes sense!

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
garyrussell: As you can see from below, I just pulled these up from the FTP config.

It would be useful if we could get the final doc merge stuff in place ASAP so we don't have to redo all this stuff.

garyrussell: Oh - looks like we're using the META-INF ...

https://github.com/spring-cloud/spring-cloud-stream-modules/pull/44/files

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryProperties.java
garyrussell, line 52: Cool

garyrussell, line 52: Except I don't think it applies here because it's not a runtime expression, it's a context initialization expression.

common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
garyrussell: Yes, I am not at all a fan of `@Test(expected=` - I prefer to do further evaluation of the exception to verify it's the exact one I wanted.

common/spring-cloud-stream-modules-test-support/src/main/java/org/springframework/cloud/stream/modules/test/file/remote/TestSftpServer.java
garyrussell, line 47: I had to do a bunch of rework here in the next PR (FTP sink) so I could reuse it for the sink.

Let's hold off from making more changes here until the next PR.

ftp-source/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSourceIntegrationTests.java
garyrussell, line 110: How can you use that to inject dynamic properties, such as `ftpServer.getPort()` ?

garyrussell, line 110: Ah - I see you realized that further down.

sftp-source/src/test/resources/logback.xml
garyrussell: Again; I started with the existing FTP source, which has one of these.

I'll remove both.

garyrussell:
Pushed - I think I got everything (except reworking the (S)FTP test servers, which I'll do in the next PR for the FTP sink).

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/sftp/SftpSessionFactoryProperties.java
ericbottard, line 52: Doesn't matter. What I'm saying is that it would surface a parsing exception as a binder exception (hence appearing alongside other user PEBCAK errors), instead of a context initialization error.
Doesn't matter much if it does not though

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,161 @@
...
+	public boolean isPreserveTimestamp() {
+		return preserveTimestamp;
+	}
+
+	public void setPreserveTimestamp(boolean preserveTimestamp) {
+		this.preserveTimestamp = preserveTimestamp;
+	}
+
+	@AssertTrue(message = "filenamePattern and filenameRegex are mutually exclusive")
+	public boolean exclusivePatterns() {
```
ericbottard, line 157: This won't work. Needs to be a javabean style getter (hence `isExclusivePatterns()` or `getExclusivePatterns()`)

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
ericbottard: Can't find this comment anymore, maybe you removed it Gary.
To clarify, yes we're using the stuff in META-INF, but that is generated in
99% of the cases (hence my comments about needing proper javadoc, no need
to document the default, etc).
The file you pointed to in my other PR (#44) is needed because
`pathPattern` there is not on a `@ConfigurationProperties` class, hence the
need to write the json file "by hand"
Hope that makes sense

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileSourceProperties.java
```diff
@@ -0,0 +1,161 @@
...
+
+	/**
+	 * Set to true to delete remote files after successful transfer.
+	 */
+	private boolean deleteRemoteFiles = false;
+
+	/**
+	 * The local directory to use for file transfers.
+	 */
+	private File localDir = new File(System.getProperty("java.io.tmpdir") + "/xd/ftp");
```
ericbottard, line 48: Need to remove reference to XD here

ericbottard:
Merged as b0ab24301cd64d172a36ae3f4cbbe614889dfe02 with f56cb9dfcbee61df17d0b24467d8dcc435211cd2 as an additional commit, which we may want to revisit as a better way to handle `withMarkers`

PR #43, opened 2015-10-07
XD-3603: Port XD Log Sink (replace existing)

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3602

Initial s-c-s log sink is not configurable.

Make it compatible with the XD sink.

### DIFF
--- a/log-sink/pom.xml
+++ b/log-sink/pom.xml
@@ -21,6 +21,7 @@
 		<dependency>
 			<groupId>org.springframework.integration</groupId>
 			<artifactId>spring-integration-test</artifactId>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud.stream.module</groupId>
--- a/log-sink/pom.xml
+++ b/log-sink/pom.xml
@@ -17,4 +17,15 @@
 		<start-class>org.springframework.cloud.stream.module.log.LogSinkApplication</start-class>
 	</properties>
 
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-test</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud.stream.module</groupId>
+			<artifactId>spring-cloud-stream-modules-test-support</artifactId>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
 </project>
--- a/log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSink.java
+++ b/log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSink.java
@@ -16,25 +16,33 @@
 
 package org.springframework.cloud.stream.module.log;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.messaging.Sink;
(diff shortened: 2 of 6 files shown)


log-sink/pom.xml
```diff
@@ -17,4 +17,15 @@
 		<start-class>org.springframework.cloud.stream.module.log.LogSinkApplication</start-class>
 	</properties>
 
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-test</artifactId>
```
markfisher, line 23: this should be test scope?

log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSink.java
```diff
@@ -16,25 +16,33 @@
...
  * @author Marius Bogoevici
+ * @author Gary Russell
  */
 @EnableBinding(Sink.class)
+@EnableConfigurationProperties(LogSinkProperties.class)
 public class LogSink {
 
-	private static Logger logger = LoggerFactory.getLogger(LogSink.class);
+	@Autowired
+	private LogSinkProperties config;
```
markfisher: we're still not 100% consistent across modules, but in most cases I think we're referring to the properties instance as `properties`

log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSinkProperties.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Gary Russell
+ *
+ */
+@ConfigurationProperties
+public class LogSinkProperties {
+
+	private String name = "log.sink"; // TODO: set from environment 'group' if available
```
markfisher: in other cases, we've used `${spring.application.name}` (with a default like this value) for now, e.g.:

https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/hdfs-sink/src/main/java/org/springframework/cloud/stream/module/hdfs/sink/HdfsSinkProperties.java#L47

log-sink/pom.xml
garyrussell, line 23: oops; yes - it was automatically added by eclipse.

log-sink/src/main/java/org/springframework/cloud/stream/module/log/LogSink.java
garyrussell: I guess I picked a bad template :smile:  (FTP)

garyrussell:
Pushed

markfisher:
LGTM; merging

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/ea2c770e5f4580eeecf1648d898f722744d98a74

log-sink/src/test/java/org/springframework/cloud/stream/module/log/LogSinkApplicationTests.java
```diff
@@ -39,11 +52,39 @@
...
 
+	@SuppressWarnings("unused")
 	@Autowired
 	private Sink same;
 
+	@Autowired
+	private LoggingHandler logSinkHandler;
+
+	@BeforeClass
+	public static void setUp() {
```
ericbottard, line 63: Sorry, just looking at this now.
This `PropertiesInitializer` stuff should be replaced by `@[Web]IntegrationText({"name=foo", etc...})` IMHO

garyrussell, line 63: I was following the existing model in the FTP source.

If that model is incorrect, it should be fixed before asking "newbies" to contribute :smile:

PR #45, opened 2015-10-14
Added Gemfire repo

pperalta:


trisberg:
Why is this necessary?

pperalta:
Without it, @markfisher and I get this:

> [ERROR] Failed to execute goal on project spring-cloud-stream-modules-test-support: Could not resolve dependencies for project org.springframework.cloud.stream.module:spring-cloud-stream-modules-test-support:jar:1.0.0.BUILD-SNAPSHOT: Failed to collect dependencies at com.gemstone.gemfire:gemfire:jar:8.0.0: Failed to read artifact descriptor for com.gemstone.gemfire:gemfire:jar:8.0.0: Could not transfer artifact com.gemstone.gemfire:gemfire:pom:8.0.0 from/to spring-snapshots (http://repo.spring.io/libs-snapshot-local): Failed to transfer file: http://repo.spring.io/libs-snapshot-local/com/gemstone/gemfire/gemfire/8.0.0/gemfire-8.0.0.pom. Return code is: 409, ReasonPhrase: Conflict. -> [Help 1]

trisberg:
Strange, I just built the project and it resolved all the gemfire jars fine. 

Can you try changing `<url>http://repo.spring.io/libs-snapshot-local</url>` to be `<url>http://repo.spring.io/snapshot</url>` in `.settings.xml`

trisberg:
This artifact should really be resolved via `<url>http://repo.spring.io/libs-release</url>` not sure why the snapshot lib pulled it in. Maybe we should put the `libs-release` repo first?

pperalta:
I just tried it again, and this time it executed just fine without any changes. 😕

Could it be a problem with the repo itself?

trisberg:
I think the repo did the right thing, its not a snapshot release so it shouldn't resolve using a snapshot repo reference. This is the error I get when accessing it using the link above:

status: 409,
message: "The repository 'libs-snapshot-local' rejected the artifact 'libs-snapshot-local:com/gemstone/gemfire/gemfire/8.0.0/gemfire-8.0.0.pom' due to its snapshot/release handling policy."

pperalta:
In that case, should we just move the releases repo to the top of the list?

FWIW, this also worked:

```
diff --git a/.settings.xml b/.settings.xml
index aad3442..74904ae 100644
--- a/.settings.xml
+++ b/.settings.xml
@@ -17,7 +17,7 @@
                                <repository>
                                        <id>spring-snapshots</id>
                                        <name>Spring Snapshots</name>
-                                       <url>http://repo.spring.io/libs-snapshot-local</url>
+                                       <url>http://repo.spring.io/snapshot</url>
                                        <snapshots>
                                                <enabled>true</enabled>
                                        </snapshots>
```

trisberg:
Hard to tell unless we can reliably reproduce the issue. Don't think moving the release repo to the top would hurt. Also not sure why we use the full internal repo name instead of just `http://repo.spring.io/milestone` and `http://repo.spring.io/snapshot`.

mbogoevici:
Not entirely clear why we don't simply use `http://repo.spring.io/libs-snapshot`, which aggregates them all (including the gemstone repo). I'd avoid proliferating the list of repositories as much as possible.

markpollack:
Just tried Marius' suggestion and it looks to work.  Not sure why the spring-cloud project doesn't follow this as well.  Let's ping those folks.

pperalta:
Closed in favor of https://github.com/spring-cloud/spring-cloud-stream-modules/pull/50

markpollack:
So backtracking here.  This PR looks to be the better approach vs. #50.  The reason is that if you include `http://repo.spring.io/libs-snapshot` then you get spring project dependencies and all their transitive dependencies hosted by `libs-snapshot`, sort of a wildcard.  If one uses the more selective `-local` flavored repos you don't pick up any other transitive dependencies other than those in maven central.  That keeps the knowledge of what extra repos are needed more explicit.  The core issue is that gemfire libraries are not in maven central, so need to be referenced some how.  

I'll merge this with an additional comment that the gemfire libs are not in maven central and also remove the `<servers/>` settings as we use Bamboo+Artifactory for the release and those values will be provided via the Artifactory Bamboo plugin.

markpollack:
Merged with a few other minor changes as e157b4bf0d7c5ef1c7f705ece18b0fe6f95663a9

PR #46, opened 2015-10-14
XD-3571 Add Cassandra sink module

trisberg:
- Ported existing Spring XD Cassandra sink module


cassandra-sink/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+					<groupId>org.slf4j</groupId>
+					<artifactId>slf4j-log4j12</artifactId>
+				</exclusion>
+			</exclusions>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<scope>test</scope>
```
artembilan: I wonder if we can get rid of this dependency for that tiny `Supplier` usage and just switch to Java 8 for tests.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * @author Artem Bilan
+ * @author Thomas Risberg
+ */
+@Configuration
+@EnableConfigurationProperties(CassandraProperties.class)
+public class CassandraConfiguration extends AbstractCassandraConfiguration {
+
+	@Autowired
+	CassandraProperties cassandraSinkProperties;
```
artembilan: Bad name for the property.
My idea was to extract some common options when we will add the Cassandra Source module.
That's why we have that specific `CassandraSinkProperties`.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.util.StdDateFormat;
+
+/**
+ * @author Artem Bilan
+ * @author Thomas Risberg
+ */
+@Configuration
+@Import(CassandraConfiguration.class)
+@EnableIntegration
```
artembilan: Having `@SpringBootApplication` this is redundant.
Minor, of course...

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+	private static final SpelExpressionParser PARSER = new SpelExpressionParser();
+
+	@Autowired
+	private CassandraSinkProperties cassandraSinkProperties;
+	
+	@Autowired
+	private CassandraOperations template;
+
+	@Bean
+	public MessageChannel input() {
```
artembilan: This isn't in use any more.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSink.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+@EnableBinding(Sink.class)
+public class CassandraSink {
+
+	@Autowired
+	private MessageHandler cassandraSinkMessageHandler;
+
+	@ServiceActivator(inputChannel=Sink.INPUT)
+	public void cassandraSink(Message<?> message) {
+		cassandraSinkMessageHandler.handleMessage(message);
```
artembilan: Looking to this class and how it is wired with the `CassandraSinkConfiguration`, I'd call it as an overhead.
How about rename the `CassandraSinkConfiguration` to this name, move `@EnableBinding(Sink.class)` there, as well as this `@ServiceActivator` annotation?
Otherwise we end up with extra `handleMessage()`...

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
```diff
@@ -0,0 +1,200 @@
...
+	private static final int PORT = 9043; // See spring-cassandra.yaml - native_transport_port
+
+    @Autowired
+    @Bindings(CassandraSink.class)
+    protected Sink sink;
+
+    @Autowired
+    ConfigurableApplicationContext applicationContext;
+
+	private static Cluster cluster;
```
artembilan, line 78: One of us has wrong indents confing...
Tabs instead of whitespaces. 
Or have I missed anything already?..

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
```diff
@@ -0,0 +1,200 @@
...
+	@AfterClass
+	public static void cleanup() {
+		if (cluster != null) {
+			cluster.close();
+		}
+		EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
+	}
+
+	@WebIntegrationTest({"spring.cassandra.port=9043", 
+		"spring.cassandra.keyspace=cassandraTest",
```
artembilan: Can't we really use here constants defined above ?

cassandra-sink/src/test/resources/cassandra.properties
```diff
@@ -0,0 +1,4 @@
+info.shortDescription = Ingests data into a Cassandra table.
+options_class = org.springframework.xd.cassandra.CassandraSinkOptionsMetadata
+base_packages = org.springframework.xd.cassandra
```
artembilan: What is the purpose of this `cassandra.properties` file, especially in the `test` scope ?

cassandra-sink/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+				<exclusion>
+					<groupId>com.datastax.cassandra</groupId>
+					<artifactId>cassandra-driver-core</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
+					<artifactId>slf4j-log4j12</artifactId>
+				</exclusion>
+			</exclusions>
+			<scope>test</scope>
```
ericbottard: Maybe move this scope element before the exclusions.
Also, a comment explaining why there are exclusions would be welcome

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+	
+	@Autowired
+	private Session session;
+
+	@PostConstruct
+	public void init() throws IOException {
+		if (cassandraSinkProperties.getInitScript() != null) {
+			Resource initScriptResource = new DefaultResourceLoader().getResource(cassandraSinkProperties.getInitScript());
```
ericbottard: Can make the property on the `cassandraSinkProperties` object a `Resource` directly

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+			Scanner scanner = new Scanner(initScriptResource.getInputStream(), "UTF-8");
+			String scripts = null;
+			try {
+				scripts = scanner.useDelimiter("\\A").next();
+			}
+			finally {
+				if (scanner != null) {
+					scanner.close();
```
ericbottard: try with resource is your friend here :)

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
```diff
@@ -0,0 +1,162 @@
...
+ * Properties for the cassandra sink.
+ *
+ * @author Artem Bilan
+ * @author Thomas Risberg
+ */
+@ConfigurationProperties("spring.cassandra")
+public class CassandraProperties {
+
+	/*
+	 * The comma-delimited string of the hosts to connect to Cassandra
```
ericbottard: All javadoc comments need dots at the end.

ericbottard: And need to be actual javadoc, not regular comments

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
```diff
@@ -0,0 +1,162 @@
...
+	public void setMetricsEnabled(boolean metricsEnabled) {
+		this.metricsEnabled = metricsEnabled;
+	}
+
+	@NotNull
+	public String getContactPoints() {
+		return this.contactPoints;
+	}
+
+	public int getPort() {
```
ericbottard, line 126: can add range 0-65535 validation

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+		cassandraMessageHandler.setProducesReply(false);
+		if (cassandraSinkProperties.getConsistencyLevel() != null || cassandraSinkProperties.getRetryPolicy() != null || cassandraSinkProperties.getTtl() > 0) {
+			cassandraMessageHandler.setWriteOptions(
+					new WriteOptions(cassandraSinkProperties.getConsistencyLevel(), cassandraSinkProperties.getRetryPolicy(), cassandraSinkProperties.getTtl()));
+		}
+		if (StringUtils.hasText(cassandraSinkProperties.getIngestQuery())) {
+			cassandraMessageHandler.setIngestQuery(cassandraSinkProperties.getIngestQuery());
+		}
+		else if (StringUtils.hasText(cassandraSinkProperties.getStatementExpression())) {
+			Expression expression = PARSER.parseExpression(cassandraSinkProperties.getStatementExpression());
```
ericbottard: Can make the field on the properties object an `Expression` directly, provided you add `SpelExpressionConverterConfiguration` see https://github.com/spring-cloud/spring-cloud-stream-modules/blob/b12eaba454a202a91668ee4dd4f7518bebed3616/transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/TransformProcessor.java#L35

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+			}
+		}
+
+	}
+
+	@SuppressWarnings("serial")
+	private static class ISO8601StdDateFormat extends StdDateFormat {
+
+		@Override
+		protected boolean looksLikeISO8601(String dateStr) {
```
ericbottard, line 197: What's the point of this?

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+				List<Map<String, Object>> model = this.jsonObjectMapper.fromJson(payload, List.class);
+				List<List<Object>> data = new ArrayList<>(model.size());
+				for (Map<String, Object> entity : model) {
+					List<Object> row = new ArrayList<>(this.columns.size());
+					for (String column : this.columns) {
+						Object value = entity.get(column);
+						if (value instanceof String) {
+							String string = (String) value;
+							if (this.dateFormat.looksLikeISO8601(string)) {
+								synchronized (this.dateFormat) {
```
ericbottard, line 172: why not instantiate one per method invocation really?

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * @author Artem Bilan
+ * @author Thomas Risberg
+ */
+@ConfigurationProperties
+public class CassandraSinkProperties {
+
+	/*
+	 * The consistencyLevel option of WriteOptions
```
ericbottard: Dots at end, need real javadoc style comments (not regular comments)

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
```diff
@@ -0,0 +1,109 @@
...
+	 */
+	private CassandraMessageHandler.Type queryType;
+
+	/*
+	 * The ingest Cassandra query
+	 */
+	private String ingestQuery;
+
+	/*
+	 * The expression in Cassandra query DSL style
```
ericbottard: Says "in Cassandra query DSL style", yet this is parsed as SpEL o.O

See my earlier comment about making this a SpEL `Expression` object directly then, if applicable

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
```diff
@@ -0,0 +1,200 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = CassandraSinkApplication.class)
+@DirtiesContext
+public abstract class CassandraSinkIntegrationTests {
+
+	public static final String CASSANDRA_KEYSPACE = "cassandraTest";
+
+	private static final String CASSANDRA_STORAGE_CONFIG = "spring-cassandra.yaml";
+
+	private static final int PORT = 9043; // See spring-cassandra.yaml - native_transport_port
```
ericbottard: This should be made dynamic and find an available port

cassandra-sink/pom.xml
ericbottard: Could even port our EventuallyMatcher from XD, which is more versatile than just `assertEqualsEventually`

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
ericbottard: Can.

Can't use a dynamic port though :(

cassandra-sink/pom.xml
ericbottard: Hmm, on second thoughts, not sure it works exactly the same way. You need repetitive evaluation of the template here.
In all cases, dragging reactor for just a single callback interface indeed feels overhead.

cassandra-sink/src/test/resources/cassandra.properties
ericbottard: This is XD stuff, needs to go.

cassandra-sink/src/test/resources/int-db.cql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS book;
```
ericbottard: should this be named in_i_t-db.cql ?

cassandra-sink/src/test/resources/logback.xml
```diff
@@ -0,0 +1,10 @@
+<configuration>
```
ericbottard, line 2: this should be removed

cassandra-sink/pom.xml
artembilan: When I was introducing that `assert` for the XD Modules project the Reactor dependency was there automatically from some XD dependency.
Since we have to specify it here explicitly, I'm sure we can remove it in favor of some other solution. And even if we can't use Java 8 from here, we can just introduce that simple `Supplier` interface as an inner class definition exactly in that specific test class.

ericbottard: Absolutely

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
artembilan: ???
Not sure what you mean...
I even don't understand right now why I use here `scanner`...
But anyway I don't see difference.
We read the whole file to string and then delimit it to several scripts.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
artembilan, line 197: ???
Please, be more specific.
1. 

```
/**
  * Defines a commonly used date format that conforms
  * to ISO-8601 date formatting standard, when it includes basic undecorated
  * timezone definition
*/
protected final static String DATE_FORMAT_STR_ISO8601 = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
```

You can kill me, but I don't remember already why I have chosen this format...
Maybe it is some String format which can be inserted as is to the Cassandra Date column?..
http://docs.datastax.com/en/cql/3.0/cql/cql_reference/timestamp_type_r.html 
1. We need an extension for this class to have access to this `boolean` method. See the logic in the `transformPayload()` above.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
ericbottard: I'm saying you can do 

```
try (Scanner scanner = new Scanner(...)) {
   scripts = scanner.useDelimiter("\\A").next();
}
```

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
ericbottard, line 197: My point was: why override a method just to call `super` ?

ericbottard, line 197: > We need an extension for this class to have access to this boolean method. See the logic in the transformPayload() above.

Ah, I think I see. This makes the method visible. Maybe we should add a comment to explain this then

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
artembilan: Doh... Haven't paid attention to the Java 7 features yet...
Thank you ! :smile:

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
artembilan, line 172: Let me guess: because GC is more expensive than short `synchronized`.
Pay attention that its field is `final`, so it isn't counted by GC at all.

cassandra-sink/src/test/resources/int-db.cql
artembilan: +1.
Although it doesn't matter because it is the test scope resource.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
trisberg: Yes, will rename it

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
trisberg: Right, forgot to remove it

cassandra-sink/pom.xml
trisberg: Sure, @artembilan any notes on why these exclusions are necessary, I'm assuming it is because the `cassandra-unit-spring` dependency is older and bring sin outdated artifacts plus the usual logging problems.

artembilan: Yes, that's the best reason.
And maybe that's why guys don't comment it even on the core project: https://github.com/spring-projects/spring-data-cassandra/blob/master/pom.xml#L149

cassandra-sink/pom.xml
trisberg: Agree, I'm adding an inner class interface def and removing reactor dependency

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
trisberg: Yep, less code is better

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
trisberg, line 78: must have cut'n pasted this code, should be tabs, fixing

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
trisberg: switching to use constants. For dynamic ports, not even sure that is possible with Cassandra, There are quite a few ports required in the yml file, not sure if we could set system properties at  runtime to override them. I did change a few unneeded ones to use '0'.

cassandra-sink/src/test/resources/cassandra.properties
trisberg: Oops, got copied over, removing.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
trisberg: making this change

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
trisberg: fixing

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
trisberg, line 126: yep, adding

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
trisberg, line 197: adding a comment

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
trisberg: fixing

cassandra-sink/src/test/resources/logback.xml
trisberg, line 2: gone

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
trisberg: Are they mutually exclusive?

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
```diff
@@ -0,0 +1,109 @@
...
+
+	/**
+	 * The ingest Cassandra query.
+	 */
+	private String ingestQuery;
+
+	/**
+	 * The expression in Cassandra query DSL style.
+	 */
+	private String statementExpression;
```
artembilan: > Says "in Cassandra query DSL style", yet this is parsed as SpEL o.O

My bad. Not sure how much letters you allow for such a comment on properties, but my intention was to say here that is:
A SpEL to be evaluated to the `com.datastax.driver.core.Statement` where Cassandra query DSL can be used without FQCN. E.g. 

```
--statementExpression=QueryBuilder.select('c1','c2').from('t').where(QueryBuilder.eq('c3', payload))
```

> Are they mutually exclusive?

If you mean together with `ingestQuery`, than yes. they are. But the `ingestQuery` has precedence in the `CassandraMessageHandler`. All other options are just ignored.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSink.java
markfisher: +1 but we can even keep the name `CassandraSinkconfiguration` since there is no requirement to have a class that is _the_ Sink.

trisberg:
@artembilan I was going to update this PR but the integration tests no longer pass. Get this error - com.datastax.driver.core.exceptions.InvalidQueryException: Keyspace 'cassandraTest' does not exist. To me it looks like we create that when the test class is loaded. Any ideas?

artembilan:
@trisberg , would you mind sharing your code to play from my side locally?

Thanks

artembilan:
Well, it doesn't pass for me even without your polishing.
Investigating....

artembilan:
Not sure what have been changed, but it can't connect because the `keyspace` must be quoted:
http://log.damnpeople.fr/2014/10/cassandra-keyspace-case-sensitivity-wtf/
http://stackoverflow.com/questions/29569443/cassandra-keyspace-name-with-hypen

So, we should just fix our `cassandraTest` name to something like `cassandratest` or just `sink`.
But... That doesn't help me to pass the test. I'm finishing with something like JVM crash. 
Looks like anything has been changed recently in the SF which tells us that we aren't correct.

cassandra-sink/src/test/resources/spring-cassandra.yaml
```diff
@@ -0,0 +1,630 @@
...
+#
+# See http://wiki.apache.org/cassandra/Operations for more on
+# partitioners and token selection.
+partitioner: org.apache.cassandra.dht.Murmur3Partitioner
+
+# Directories where Cassandra should store data on disk.  Cassandra
+# will spread data evenly across them, subject to the granularity of
+# the configured compaction strategy.
+data_file_directories:
+    - build/embeddedCassandra/data
```
artembilan: s/build/target
just because we are with Maven now, not Gradle...

cassandra-sink/src/test/resources/spring-cassandra.yaml
```diff
@@ -0,0 +1,630 @@
...
+
+# commit log
+commitlog_directory: build/embeddedCassandra/commitlog
```
artembilan: DITO

cassandra-sink/src/test/resources/spring-cassandra.yaml
```diff
@@ -0,0 +1,630 @@
...
+#
+# Default is 0 to disable saving the row cache.
+row_cache_save_period: 0
+
+# Number of keys from the row cache to save
+# Disabled by default, meaning all keys are going to be saved
+# row_cache_keys_to_save: 100
+
+# saved caches
+saved_caches_directory: build/embeddedCassandra/saved_caches
```
artembilan: DITO

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * @author Artem Bilan
+ * @author Thomas Risberg
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = CassandraSinkApplication.class)
+@DirtiesContext
+public abstract class CassandraSinkIntegrationTests {
+
+	public static final String CASSANDRA_KEYSPACE = "cassandraTest";
```
artembilan: So, as we talked before this should be changed to something only in lower case.
Just because CQL doesn't care about quotes.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
```diff
@@ -0,0 +1,133 @@
...
+public class CassandraConfiguration extends AbstractCassandraConfiguration {
+
+	@Autowired
+	CassandraProperties cassandraProperties;
+	
+	@Autowired
+	private Session session;
+
+	@PostConstruct
+	public void init() throws IOException {
```
artembilan: Looks like we aren't correct here and introduce the circular dependency:
`session` is created by the `@Bean CassandraSessionFactoryBean` in the superclass.

Right, when I comment `@Autowired` on the `session` and `@PostConstruct` the test goes a bit further.

Investigating what's going on here with `Invalid version for TimeUUID type`.

Meanwhile I'd ask you for the best place to call init scripts for the DB before the start of our adapters, of course...

artembilan: I hacked it with moving all this initialization to the inner class:

```
@Configuration
protected static class CassandraInitializerConfiguration {

 }
```

After that it started to pass for me - the whole `cassandra-sink` module.
It is very long testing process though with a lot of `DEBUG` in logs, but it passes.

So, @trisberg , feel free to push your changes and than I'll share with you my polishing. 
Or you can address it yourself already, of course.

trisberg:
Think I solved the test issues. Rebased the code.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * The port to use to connect to the Cassandra host.
+	 */
+	private int port = CassandraCqlClusterFactoryBean.DEFAULT_PORT;
+
+	/**
+	 * The keyspace name to connect to.
+	 */
+	@Value("${spring.application.name:tmp}")
```
artembilan, line 51: Is the `tmp` intentional?

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
artembilan: Didn't we find the compromise with this `Expression` stuff ?
At least, shouldn't we follow here with the `FileSinkConfiguration` experience ?

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkApplication.java
```diff
@@ -0,0 +1,31 @@
...
+package org.springframework.cloud.stream.module.cassandra.sink;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+/**
+ * @author Thomas Risberg
+ */
+@SpringBootApplication
+public class CassandraSinkApplication {
```
artembilan, line 26: It's interesting if we need this separate class at all.
Looks like it is the same in each module, which are the independent applications at all. 
So, no need to worry about the name of class.

We even could generate it... Or try to figure out some other way to bypass this boring copy/paste work...

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
```diff
@@ -0,0 +1,204 @@
...
+
+	
+	private static List<Book> getBookList(int numBooks) {
+
+		List<Book> books = new ArrayList<>();
+
+		Book b;
+		for (int i = 0; i < numBooks; i++) {
+			b = new Book();
+			b.setIsbn(UUID.randomUUID());
```
artembilan, line 178: What is interesting that this test data is the same like for the `insert` test.
From other side we have the table script:

```
isbn        uuid PRIMARY KEY,
```

not `timeuuid`, like it barks to us for the `insert` test...

artembilan:
The top comment.
You should configure your IDE to remove trailing whitespaces. Otherwise a lot of auto-polishing from my side :smile:

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
trisberg, line 51: yes, other sinks tend yo use tmp or /tmp as a default

trisberg:
I do have that configured, but only for "Modified Lines", not sure where they came from, maybe the original files I copied.

artembilan:
Well, I use "All" mode to fix an existing code during PRs.

cassandra-sink/src/test/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkIntegrationTests.java
artembilan, line 178: I know "why?": you had `"spring.cassandra.init-script=init-db.cql` before for this test, too.
But right now it is based on the Spring Data Entity Mapping mechanism.
And we end up with the

```
TIMEUUID  (15, UUID.class),
```

for `Book.isbn`.

So, that your `book.setIsbn(UUIDs.timeBased());` is a good fix.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraConfiguration.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+@Configuration
+@EnableConfigurationProperties(CassandraProperties.class)
+public class CassandraConfiguration extends AbstractCassandraConfiguration {
+
+	@Autowired
+	CassandraProperties cassandraProperties;
+
+	@Override
+	protected List<String> getStartupScripts() {
```
artembilan: Let me replicate my comment from commit here one more time.

I didn't use this approach from the beginning because this getStartupScripts() is used for Cluster configuration as well as for Session.
See the usage of this method.

My intention of these init scripts was only to initialize the specific keyspace. And only session is for that.
I admit that it looks like a bug in the Spring Data Cassandra, because we are sharing the same scripts for the whole Cluster and the specific session for particular keyspace.
Maybe I don't know Cassandra so well, but I'm afraid that we may have some side effect to go this strange way...

artembilan: As I said it is dangerous. Scripts are executed on the `system` keyspace - `CassandraCqlClusterFactoryBean`:

```
if (system == null) {
                    system = scripts.size() == 0 ? null : cluster.connect();
                }

                if (template == null) {
                    template = system == null ? null : new CqlTemplate(system);
                }

                for (String script : scripts) {

                    if (log.isDebugEnabled()) {
                        log.debug("executing raw CQL [{}]", script);
                    }

                    template.execute(script);
}
```

So, in our case we create the same table in two keyspaces.
Fortunately with Embedded Cassandra we have all those rights.
With the real Cassandra need to test.
But, at least, I sent you a bell here...

artembilan: I have it locally with inner class like this:

``` java
public class CassandraConfiguration extends AbstractCassandraConfiguration {

    @Autowired
    CassandraProperties cassandraProperties;

    /**
     * The inner class to execute init scripts on the provided {@code keyspace}.
     * It is here to bypass circular dependency with {@link Session} injection and
     * its {@code @amp;Bean} in the {@link AbstractCassandraConfiguration}.
     * From other side we can't use {@link #getStartupScripts()} because they are
     * shared between Cluster and Session initializations.
     */
    @Configuration
    protected static class CassandraInitializerConfiguration {

        @Autowired
        CassandraProperties cassandraProperties;

        @Autowired
        private Session session;

        @PostConstruct
        public void init() throws IOException {
            if (cassandraProperties.getInitScript() != null) {
                Resource initScriptResource = new DefaultResourceLoader().getResource(cassandraProperties.getInitScript());
.....
```

cassandra-sink/src/test/resources/logback.xml
artembilan, line 2: ??? hasn't gone yet...

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
artembilan: This isn't addressed yet and even without comments...

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkProperties.java
artembilan: If you'd like to see some other comment on this property, please, count my comment here.

cassandra-sink/src/test/resources/logback.xml
trisberg, line 2: it was gone for a while, added it back in to get rid of all the logging during @BeforeClass

artembilan:
Everything else for me looks good here.

cassandra-sink/src/test/resources/logback.xml
artembilan, line 2: Yes, that's why I ask in the `s-c-s` Slack channel:

> @here, can you tell me where is the `logging` config for the `s-c-s-modules` and why so much `DEBUG` noise do you allow? Thanks

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/sink/CassandraSinkConfiguration.java
trisberg: pushed new fix to address this

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
```diff
@@ -0,0 +1,164 @@
...
+
+import org.hibernate.validator.constraints.Range;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cassandra.config.CassandraCqlClusterFactoryBean;
+import org.springframework.cassandra.config.CompressionType;
+import org.springframework.util.StringUtils;
+
+/**
+ * Properties for the cassandra sink.
```
artembilan, line 30: This is common properties. Not only for the `sink`.
Minor: will be fixed on merge.

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * The password for connection.
+	 */
+	private String password;
+
+	/**
+	 * The path to file with CQL scripts (delimited by ';') to initialize keyspace schema.
+	 */
+	private String initScript;
```
artembilan, line 67: I found that this one can be specified as a `Resource`.
Then in the debug I see this result:

```
result = {DefaultResourceLoader$ClassPathContextResource@7696} "class path resource [init-db.cql]"
 path = "init-db.cql"
 classLoader = {Launcher$AppClassLoader@7699} 
 clazz = null
```

Let me know if we can go ahead with the `Resource` approach and I'll fix it on merge.
Otherwise explain, please, what is the concern exists do not do that.
Thanks

ericbottard, line 67: I'd vote for `Resource` type as well, and let the converter do the resolve

artembilan:
Here is a polishing for your confirmation: https://github.com/artembilan/spring-cloud-stream-modules/commit/b1404ac688ba26d225d55e9c19f9418ebb2ae782.
Tested against "real" Cassandra, although it doesn't matter: the Cassandra is just a Java application :smile:

cassandra-sink/src/main/java/org/springframework/cloud/stream/module/cassandra/CassandraProperties.java
trisberg, line 67: Resource sounds good

trisberg:
Looks OK. One comment, it seems you use the `this.` prefix everywhere while most other modules do not. Seems to be an SI thing, don't think most other Spring projects use that style.

artembilan:
No, that is really SF code style on which we rely in SI: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style:

> A field of a class should always be referenced using this. A method of class, however, should never be referenced using this.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/29779f033bd38d62cbe5e5161621cff7af8170cf

PR #47, opened 2015-10-15
Remove some toxic dependencies from gemfire

ericbottard:
We don't really need those, and the mess the build in IDEA.


markpollack:
merged 8aa294c5597b98aef841067ff17205a289c15595

PR #48, opened 2015-10-15
XD-3370: FTP Sink

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3370

Refactor common source/sink properties for remote file source/sink.

Refactor test servers for use in both source and sink and to properly
use JUnit `TemporaryFolder`.

### DIFF
--- a/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileProperties.java
+++ b/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileProperties.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.module.file.remote;
+
+import org.hibernate.validator.constraints.NotBlank;
+
+/**
+ * @author Gary Russell
+ *
+ */
+public abstract class AbstractRemoteFileProperties {
+
+	/**
+	 * The remote FTP directory, "/" by default.
+	 */
+	private String remoteDir = "/";
+
+	/**
+	 * The suffix to use while the transfer is in progress, ".tmp" by default.
+	 */
+	private String tmpFileSuffix = ".tmp";
+
+	/**
+	 * The remote file separator, "/" by default.
+	 */
+	private String remoteFileSeparator = "/";
+
+	@NotBlank
+	public String getRemoteDir() {
+		return remoteDir;
+	}
+
+	public void setRemoteDir(String remoteDir) {
+		this.remoteDir = remoteDir;
+	}
+
+	@NotBlank
+	public String getTmpFileSuffix() {
+		return tmpFileSuffix;
+	}
+
+	public void setTmpFileSuffix(String tmpFileSuffix) {
+		this.tmpFileSuffix = tmpFileSuffix;
+	}
+
+	@NotBlank
+	public String getRemoteFileSeparator() {
+		return remoteFileSeparator;
+	}
+
+	public void setRemoteFileSeparator(String remoteFileSeparator) {
+		this.remoteFileSeparator = remoteFileSeparator;
+	}
+
+}
--- a/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileProperties.java
+++ b/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileProperties.java
@@ -24,17 +24,17 @@ import org.hibernate.validator.constraints.NotBlank;
 public abstract class AbstractRemoteFileProperties {
 
 	/**
-	 * The remote FTP directory, "/" by default.
+	 * The remote FTP directory.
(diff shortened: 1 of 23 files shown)


ftp-sink/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud.stream.module</groupId>
+			<artifactId>spring-cloud-stream-modules-common-configuration</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-ftp</artifactId>
+			<version>${spring-integration.version}</version>
```
artembilan: Isn't version provided by the `parent`?
I guess yes, because you don't use `<version>` for the `org.springframework.integration`

ftp-sink/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud.stream.module</groupId>
+			<artifactId>spring-cloud-stream-modules-test-support</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.ftpserver</groupId>
+			<artifactId>ftpserver-core</artifactId>
+		</dependency>
```
artembilan: Why is without `<scope>test</scope>` ?

ftp-sink/pom.xml
garyrussell: To be honest, I just copied the ftp source pom - I didn't pay much attention to it (given my maven aversion :smile: ).

garyrussell:
Pushed some POM polish.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/file/remote/AbstractRemoteFileProperties.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.springframework.cloud.stream.module.file.remote;
+
+import org.hibernate.validator.constraints.NotBlank;
+
+/**
+ * @author Gary Russell
+ *
+ */
+public abstract class AbstractRemoteFileProperties {
```
artembilan, line 24: I like more when we mark them all with the `@ConfigurationProperties`, don't inherit and just specify in the `EnableConfigurationProperties`.
See the #46 for similar problem around the common properties.

I won't mind if it doesn't matter though...

ftp-sink/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSinkIntegrationTests.java
```diff
@@ -0,0 +1,90 @@
...
+
+		Properties properties = new Properties();
+		properties.put("remoteDir", "ftpTarget");
+		properties.put("username", "foo");
+		properties.put("password", "foo");
+		properties.put("filenamePattern", "*");
+		properties.put("port", port);
+		properties.put("mode", "FAIL");
+		properties.put("filenameExpression", "payload.name.toUpperCase()");
+		PropertiesInitializer.PROPERTIES = properties;
```
artembilan, line 61: I wonder if we can rely on the `System.properties` or even `ENV` variables...
Do you remember the [12 Factor App](http://12factor.net/) principles ?

ftp-sink/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSinkIntegrationTests.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @author David Turanski
+ * @author Marius Bogoevici
+ * @author Gary Russell
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = FtpSinkApplication.class, initializers = PropertiesInitializer.class)
+@DirtiesContext
+public class FtpSinkIntegrationTests extends TestFtpServer {
```
artembilan: This extension smells a bit.
Not sure which goal you aim here, but this architecture looks worse than before.
I see your point regarding `TemporaryFolder` `@Rule`, but maybe we can rework exactly that part and inject `TestFtpServer` as before and use its `getTargetRemoteDirectory()` and so on?..

I won't mind if you go this way anyway...

garyrussell: Maybe it's just the class name that concerns you - the parent class is not really a server; it has a server field. Perhaps it should be called `FtpTestSupport` instead?

artembilan: Yes, `FtpTestServerSupport` sounds good.

ftp-sink/src/main/java/org/springframework/cloud/stream/module/ftp/FtpSink.java
```diff
@@ -0,0 +1,78 @@
...
+	DefaultFtpSessionFactory ftpSessionFactory;
+
+	@Autowired
+	@Bindings(FtpSink.class)
+	Sink sink;
+
+	@Bean
+	public IntegrationFlow ftpInboundFlow() {
+		FtpMessageHandlerSpec handlerSpec =
+					Ftp.outboundAdapter(new FtpRemoteFileTemplate(this.ftpSessionFactory), this.properties.getMode())
```
ericbottard: Indentation seems a bit off, not unreadable though

garyrussell:
Pushed

ftp-sink/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>ftp-sink</artifactId>
+	<packaging>jar</packaging>
+	<name>ftp-sink</name>
+	<description>FTP Sink stream module</description>
+
+	<properties>
+		<start-class>org.springframework.cloud.stream.module.ftp.sink.FtpSinkApplication</start-class>
```
ericbottard, line 16: This classname is wrong. Will fix on merge

ericbottard:
Merged as 57054a0 as well as https://github.com/spring-cloud/spring-cloud-dataflow/commit/bfd05d9a042d7ad5247f617b85a710f7133bd34e on dataflow

@garyrussell don't forget to add the submodule to the parent pom

PR #50, opened 2015-10-19
Updated repositories

pperalta:
Based on the comments from:

https://github.com/spring-cloud/spring-cloud-stream-modules/pull/45

and the doc from:

https://github.com/spring-projects/spring-framework/wiki/Spring-repository-FAQ


markpollack:
Going back to the approach in #45 - see comments in there for the rational.

PR #51, opened 2015-10-22
XD-3622: File Source

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3622

Port file source from XD; add regex for pattern.


file-source/README.adoc
```diff
@@ -0,0 +1,33 @@
+= Spring Cloud Stream File Sink
```
artembilan: ??? Is something wrong here?
I wonder if we should tell here about `File Source` instead...

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+import org.springframework.integration.file.splitter.FileSplitter;
+import org.springframework.integration.file.transformer.FileToByteArrayTransformer;
+import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.scheduling.Trigger;
+import org.springframework.util.StringUtils;
+
+/**
+ * Creates a {@link FileReadingMessageSource} bean and registers it as a
+ * Inbound Channel Adapter that sends messages from the Source output channel.
```
artembilan: ... sends messages s/from/to the Source output channel.

ericbottard:
Don't forget to add this new module to the top-level `pom.xml` :)

file-source/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+	<description>File Source stream module</description>
+
+	<parent>
+		<groupId>org.springframework.cloud.stream.module</groupId>
+		<artifactId>spring-cloud-stream-modules</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<start-class>org.springframework.cloud.stream.module.file.FileSourceApplication</start-class>
```
ericbottard: wrong package

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
ericbottard: That sentence look awkward to me as well, but I think I understand Gary's "point of view" here, namely looking at the module from its exterior.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
```diff
@@ -0,0 +1,93 @@
...
+@ConfigurationProperties
+public class FileSourceProperties {
+
+	private static final String DEFAULT_DIR = System.getProperty("java.io.tmpdir") +
+			File.separator + "dataflow" + File.separator + "intput";
+
+	/**
+	 * The directory to poll for new files.
+	 */
+	private String directory = DEFAULT_DIR;
```
ericbottard, line 40: Wondering if this would make sense to turn this property to a `java.io.File` directly

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Set to true to include an {@code AcceptOnceFileListFilter}.
```
ericbottard: Will need to have a look at what the usage of `{@code ..}` implies in generated json metadata.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * A simple ant pattern to match files.
+	 */
+	private String filenamePattern;
+
+	/**
+	 * A regex pattern to match files.
+	 */
+	private String filenameRegex;
```
ericbottard: Could turn this to a `Pattern` (with appropriate Converter) to provide earlier binding error feedback

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
```diff
@@ -0,0 +1,196 @@
...
+@SpringApplicationConfiguration(classes = FileSourceApplication.class)
+@DirtiesContext
+public abstract class FileSourceTests {
+
+	private static final String TMPDIR = System.getProperty("java.io.tmpdir");
+
+	private static final String ROOT_DIR = TMPDIR + File.separator + "dataflow-tests"
+			+ File.separator + "input";
+
+	protected File atomicFileCreate(String filename) throws FileNotFoundException, IOException {
```
ericbottard, line 64: I don't understand why, if this method takes a parameter, it should be the name of the "hidden" file.

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
```diff
@@ -0,0 +1,196 @@
...
+		return fileFinal;
+	}
+
+	@Autowired
+	protected Source source;
+
+	@Autowired
+	protected MessageCollector messageCollector;
+
+	@IntegrationTest({ "directory = ${java.io.tmpdir}${file.separator}dataflow-tests${file.separator}input",
```
ericbottard, line 78: Look at this @markfisher . Gary is full of resources! I believe we can change some of out IT to use this approach

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
```diff
@@ -0,0 +1,196 @@
...
+		"fixedDelay = 100", "timeUnit = MILLISECONDS" })
+	public static class ContentPayloadTests extends FileSourceTests {
+
+		@Test
+		public void testSimpleFile() throws Exception {
+			String filename = "test.txt";
+			File file = atomicFileCreate(filename);
+			Message<?> received = messageCollector.forChannel(source.output()).poll(10, TimeUnit.SECONDS);
+			assertNotNull(received);
```
ericbottard, line 87: I think the returned **message** can never be `null` (same can't be said of payload variant methods)

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
```diff
@@ -0,0 +1,196 @@
...
+	public static class FilePayloadWithPatternTests extends FileSourceTests {
+
+		@Test
+		public void testSimpleFile() throws Exception {
+			File file = atomicFileCreate("test.txt");
+			File hidden = atomicFileCreate("test.foo");
+			Message<?> received = messageCollector.forChannel(source.output()).poll(10, TimeUnit.SECONDS);
+			assertNotNull(received);
+			assertEquals(file, received.getPayload());
+			received = messageCollector.forChannel(source.output()).poll(300, TimeUnit.MILLISECONDS);
```
ericbottard, line 170: Apparently I was wrong above. I thought timeout threw an exception. nvm

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
markfisher, line 78: Eh? I'm pretty sure he copied that idea from my `FileSinkTests`, e.g. https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/file-sink/src/test/java/org/springframework/cloud/stream/module/file/FileSinkTests.java#L54

ericbottard, line 78: Hmm, did not see that. Let's not get into a paternity war, shall we ;) Anyway, this is good (new) news to me!

markfisher, line 78: Well, of course I also copied it, so all credit to @trisberg (or whomever he copied it from):
https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/hdfs-sink/src/test/java/org/springframework/cloud/stream/module/hdfs/sink/HdfsSinkIntegrationTests.java#L54

...and don't get me wrong, I agree Gary is full of resources ;)

file-source/pom.xml
markfisher: file-sink's pom will need to be updated also

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @author Gary Russell
+ */
+@EnableBinding(Source.class)
+@Import({PeriodicTriggerConfiguration.class, SpelExpressionConverterConfiguration.class})
```
artembilan: I wonder if `SpelExpressionConverterConfiguration` can be declared implicitly and always unconditionally.
From one side it just adds an extra `Converter`. From another side it doesn't do anything with this `FileSource` `@Configuration`.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Autowired
+	MaxMessagesProperties maxMessagesProperties;
+
+	@Autowired
+	Trigger trigger;
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.trigger(trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
```
artembilan, line 70: Looks like this is redundant and can be as a part of `PeriodicTriggerConfiguration`.
As well as `MaxMessagesProperties` should be moved there.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+	private FileSourceProperties properties;
+
+	@Autowired
+	private FileConsumerProperties fileConsumerProperties;
+
+	@Autowired
+	MaxMessagesProperties maxMessagesProperties;
+
+	@Autowired
+	Trigger trigger;
```
artembilan: Is this fine that they are without `private`, not like a couple above ?

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.trigger(trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
+	}
+
+	@Autowired
+	Source source;
```
artembilan, line 74: Can some one explain me why do we need such a simple interface injection, if it doesn't have so much power only `output()`? 
Why just don't inject that `MessageChannel output` ?
Thanks

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+			.channel(source.output())
+			.get();
+	}
+
+	private IntegrationFlowBuilder enhanceFlowForReadingMode(IntegrationFlowBuilder flowBuilder) {
+		switch (this.fileConsumerProperties.getMode()) {
+		case contents:
+			flowBuilder.enrichHeaders(Collections.<String, Object>singletonMap(MessageHeaders.CONTENT_TYPE,
+					"application/octet-stream"))
+					.transform(new FileToByteArrayTransformer());
```
artembilan: I wonder if we should support `CONTENT_TYPE` header in those `AbstractFilePayloadTransformer`s directly...
Like we do with the `ObjectToJsonTransformer`.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+
+					});
+		return enhanceFlowForReadingMode(flowBuilder)
+			.channel(source.output())
+			.get();
+	}
+
+	private IntegrationFlowBuilder enhanceFlowForReadingMode(IntegrationFlowBuilder flowBuilder) {
+		switch (this.fileConsumerProperties.getMode()) {
+		case contents:
```
artembilan: Interesting, why enum const is in lower case...

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
artembilan: `Transformers.fileToByteArray()` BTW

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+			flowBuilder.enrichHeaders(Collections.<String, Object>singletonMap(MessageHeaders.CONTENT_TYPE,
+					"application/octet-stream"))
+					.transform(new FileToByteArrayTransformer());
+			break;
+		case lines:
+			flowBuilder.enrichHeaders(Collections.<String, Object>singletonMap(MessageHeaders.CONTENT_TYPE,
+					"text/plain"))
+					.split(new FileSplitter(true, this.fileConsumerProperties.getWithMarkers()), null);
```
artembilan: `.split(Files.splitter(true, this.fileConsumerProperties.getWithMarkers()))`
Not sure what the problem you have here with the second argument, but that works fine even with the `new FileSplitter()`

Minor, of course...

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
ericbottard: Yes, we thought about that with @markfisher. I'd vote for it too, maybe making it conditional on the absence of a converter for that already, for the rare case where the user would want his own.
Same for other converters that I think we could/should provide

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
artembilan, line 170: Yes, you were: the `poll()` works the same way as it is in the `java.util.concurrent.BlockingQueue` impl

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
ericbottard: because it's prettier IMO :sparkling_heart: 
Note that at time of binding, the Converter accepts any case

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
```diff
@@ -0,0 +1,128 @@
...
+	private IntegrationFlowBuilder enhanceFlowForReadingMode(IntegrationFlowBuilder flowBuilder) {
+		switch (this.fileConsumerProperties.getMode()) {
+		case contents:
+			flowBuilder.enrichHeaders(Collections.<String, Object>singletonMap(MessageHeaders.CONTENT_TYPE,
+					"application/octet-stream"))
+					.transform(new FileToByteArrayTransformer());
+			break;
+		case lines:
+			flowBuilder.enrichHeaders(Collections.<String, Object>singletonMap(MessageHeaders.CONTENT_TYPE,
+					"text/plain"))
```
ericbottard: Can we really assume text**/plain** here? As opposed to `text/?` ie just `text`

artembilan: That is what the next `FileSplitter` produces: only File lines as strings

ericbottard: What I'm saying is: we don't know that it is `plain`. Just because we split on newlines, doesn't mean that it's not, say, `text/xml`. Not a big deal, as we can later force the output type. But may be too big of an assumption

file-source/README.adoc
garyrussell: Strange - I remember changing it but somehow the changes are lost.

garyrussell: Found it; also the parent pom change etc; I did a `git add .` from within the module.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
garyrussell: Shall I remove it?

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
garyrussell: Doh; yes; I did that in (S)FTP after your suggestion.

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
garyrussell, line 64: Not sure what you mean; it's called to create `test.txt` and `test.foo` (which should not be found by the pattern tests).

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
garyrussell: Please create a 4.3 JIRA for the CT.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
garyrussell: True, with 1.1, but we're using 1.0.1 for some reason.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSource.java
garyrussell: This was a straight port from XD which also uses `text/plain`

garyrussell:
Pushed

garyrussell:
Pushed another commit to factor out common code from File/(S)FTP Sources.

Bumped DSL version to 1.1.0.

file-source/src/main/java/org/springframework/cloud/stream/module/file/source/FileSourceProperties.java
ericbottard: I will have a look at what is generated (you can too in `target/classes` after a maven build)

garyrussell: Looks like we need to remove it...

```
      "description": "Set to true to include an {@code AcceptOnceFileListFilter}.",
```

file-source/src/test/java/org/springframework/cloud/stream/module/file/source/FileSourceTests.java
ericbottard, line 64: What I'm saying is that (unless I'm mistaken), the file the caller gets in return is always named `test.txt` (from line 60).
The intermediate file, however, gets a name from the parameter. But we don't really care about that, as it is transient.

garyrussell:
Doh sorry - it's supposed to create a hidden file; bad refactoring; pushed.

ericbottard:
Merged as c378ca2

Also added https://github.com/spring-cloud/spring-cloud-dataflow/commit/3ec711445dd382fef988d28135c2db74f711b4a9

PR #52, opened 2015-10-23
Http Client Processor module

whummer:
This is a simple implementation of a http client processor module, which aims to reproduce the semantics of `http-client` (in Spring XD) for SCDF.

Example (assuming the module is registered as "http-client"):

```
stream create --name foo --definition "time --fixedDelay=10 | http-client --url='https://httpbin.org/post' --httpMethod=POST --body='{\"foo\":\"bar\"}' | log" --deploy
```

The log output should contain the POSTed data:

```
{
  "url": "https://httpbin.org/post",
  "data": "\\{\\\"foo\\\":\\\"bar\\\"\\}", 
  ...
}
```

### DIFF
--- a/httpclient-processor/src/main/java/org/springframework/cloud/stream/module/http/HttpClientProcessor.java
+++ b/httpclient-processor/src/main/java/org/springframework/cloud/stream/module/http/HttpClientProcessor.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.http;
+
+import java.net.URI;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Processor;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.RequestEntity;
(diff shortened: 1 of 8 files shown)


httpclient-processor/src/main/java/org/springframework/cloud/stream/module/http/HttpClientProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+			HttpHeaders headers = new HttpHeaders();
+			// TODO: add headers
+			HttpMethod method = HttpMethod.valueOf(properties.getHttpMethod());
+			URI uri = new URI(properties.getUrl());
+			HttpEntity<?> requestEntity = new RequestEntity<Object>(properties.getBody(), headers, method, uri);
+			ResponseEntity<String> httpResponse = restTemplate.exchange(
+					properties.getUrl(), 
+					HttpMethod.valueOf(properties.getHttpMethod()),
+					requestEntity,
+					String.class);
```
markfisher: expectedReturnType could be a configurable property

markfisher:
Thanks for the contribution! For some of the properties, we might want to consider supporting SpEL expressions that could be evaluated with the inbound Message as the root object for evaluation (I'm especially thinking of the request "body").

whummer:
updated the PR:
- add SpEL expression support for URL, headers, and request body
- allow configuration of expectedReturnType
- extend tests to cover the new functionality
- updated README

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/http/HttpClientProcessorProperties.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+/**
+ * Configuration properties for the Http Client Processor module.
+ *
+ * @author Waldemar Hummer
+ */
+@ConfigurationProperties
+public class HttpClientProcessorProperties {
```
markfisher, line 12: I think for the expression-based properties we probably still do want to support simple literals (i.e. not requiring the user to wrap literals with quotes). For example, we should have mutually exclusive properties `body` and `bodyExpression`. Sorry that was not clear when I originally suggested relying on Expressions. I can make that change while merging, though.

whummer, line 12: ok great. Anything else I can help with the merge, please let me know.

markfisher:
I'm reassigning this to @ericbottard, and he'll most likely replace this PR after making some additional changes to what I've added on top of this PR via this commit: https://github.com/markfisher/spring-cloud-stream-modules/commit/d909cc97b36db52b42b20de16dbc6c3706c7a680

ericbottard:
Closing in favor of https://github.com/spring-cloud/spring-cloud-stream-modules/pull/60

PR #54, opened 2015-10-26
websocket-sink based on Netty

olmoser:
This is a simple implementation of a websocket-sink based on Netty. We have a couple of cases where we need that (for Spring XD back then, now for SCDS), so there might be others happy to see a websocket-sink implementation already available.

It publishes each message received from the underlying transport to Websocket consumers via Netty channels.

Available options include the port for the Netty server, the number of worker threads or SSL support. Details are available in the readme.

An `Actuator` endpoint is available that shows the last Websocket messages sent and received via `/websocketsinktrace`.

A readme with setup and usage instructions is provided, along with integration tests that check single/multiple messages and websocket consumer cases.

I'm not sure what's your policy with community contributed source/sink/processor implementations, so if there is/will be another repo for community contributed implementations, I can file a PR for that repo.


ilayaperumalg:
@olmoser Thanks for the PR! Could you confirm if you have signed the [contributor agreement](https://support.springsource.com/spring_committer_signup).

websocket-sink/README.md
```diff
@@ -0,0 +1,127 @@
+## Netty based Websocket Sink for Spring Cloud Data Stream
```
ilayaperumalg: `Spring Cloud Data Stream` can be modified to read as `Spring Cloud Stream`.

websocket-sink/README.md
```diff
@@ -0,0 +1,127 @@
...
+SCDS version of [springxd-netty-websocket-sink](https://github.com/olmoser/spring-xd-netty-websocket-sink).
```
ilayaperumalg: Please avoid reference to individual github repositories for the code that gets committed into the main repo.

ilayaperumalg: The references of `SCDS` can be changed to `SCS`.

websocket-sink/README.md
```diff
@@ -0,0 +1,127 @@
...
+Based on the [Netty Websocket samples](https://github.com/netty/netty/tree/master/example/src/main/java/io/netty/example/http/websocketx/server)
+
+## Build and Install
+Go to the  `spring-cloud-data-stream-modules` root directory and run the following
```
ilayaperumalg: `spring-cloud-stream-modules`

websocket-sink/README.md
```diff
@@ -0,0 +1,127 @@
...
+Make sure that you use the provided `.settings.xml` in order to enable the required Spring Maven repos.
+
+## Usage
+As with the other SCDS modules, there is a single fat JAR that starts the sink implementation:
+
+```
+java -jar target/websocket-sink-*-exec.jar --spring.cloud.stream.bindings.input=ticktock
+```
+
+The default ist to bind the Netty based WebsocketServer to port `9292` and path `/websocket`. Those
```
ilayaperumalg: typo `ist`

websocket-sink/src/main/java/org/springframework/cloud/stream/module/websocket/sink/WebsocketSinkProperties.java
```diff
@@ -0,0 +1,65 @@
...
+import lombok.*;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * command line arguments available for {@link WebsocketSink}
+ *
+ * @author Oliver Moser
+ */
+@Data
+@ConfigurationProperties(prefix = "ws")
```
ilayaperumalg: We can possibly avoid having `prefix` for the module properties. Since we always pass the module properties with clear distinction on which module gets what properties, the prefix is an overhead I believe.

olmoser: Agreed. The idea behind the prefix was to have a clear distinction between e.g. the `server.port` and the (Netty Websocket) `ws.port`. 

If we remove the prefix I would suggest that I rename the properties, e.g. `port` -> `nettyPort` or `websocketPort`. Any preference regarding the naming?

-> ICLA: yes, just signed it!

ilayaperumalg: yeah, `websocketPort` sounds reasonable.

websocket-sink/src/main/resources/static/index.html
```diff
@@ -0,0 +1,74 @@
...
+		}
+		if (socket.readyState == WebSocket.OPEN) {
+			socket.send(message);
+		} else {
+			alert("The socket is not open.");
+		}
+	}
+</script>
+<form onsubmit="return false;">
+	<input name="url" value="wss://localhost:9292/websocket"/>
```
ilayaperumalg: This assumes the `SSL` is set to true. Since SSL is `false` by default, can we have this set to `ws://` instead of `wss://`?

ilayaperumalg:
LGTM.

websocket-sink/src/main/resources/static/index.html
olmoser: agree, that's a mistake, probably a left over from testing

ilayaperumalg:
Squashed, rebased and merged as 3f07431. 
@olmoser thank you once again for the PR!

PR #40, opened 2015-09-25
XD-3482: Add jdbc sink

ericbottard:
The module is, I believe, a "smarter" rewrite of the XD jdbc sink:
- Allows single inserts or opt-in batching support (the payload needs to be convertible to a `Collection`)
- Uses SpEL for computing the column value, using the `columns` option, which is a mapping from column name to SpEL. 
- A dedicated converter allows a shorthand configuration for the above though, where `columns=a,b` is equivalent to `columns=a:a, b:b`
- The SpEL is to be evaluated against both the whole message and `payload` in the single case, and collection elements in the batching case.
- This still allows XD1 compatible behavior, such as inserting the whole payload (the default is indeed `payload: payload.toString()`)
- Leverages Boot database support, so props like `spring.datasource.url` are to be used.

Comes with no driver OOTB for now, but can easily be added with `--includes`. For example using dataflow:

```
stream create foo --definition "http | jdbc --includes=com.h2database:h2:1.4.189 --spring_h2_console_enabled=true --initialize=true" --deploy
```

and go see the console at `/h2-console`

PR #55, opened 2015-10-28
XD-3482 Add jdbc sink

trisberg:
- based on Eric's PR #40
- change to use JdbcMessageHandler from Spring Integration


jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,179 @@
...
+		for (Map.Entry<String, String> entry : properties.getColumns().entrySet()) {
+			String value = entry.getValue();
+			columnExpressionVariations.add(entry.getKey(), spelExpressionParser.parseExpression(value));
+			if (!value.startsWith("payload")) {
+				columnExpressionVariations.add(entry.getKey(), spelExpressionParser.parseExpression("payload." + value));
+			}
+		}
+		JdbcMessageHandler jdbcMessageHandler = new JdbcMessageHandler(dataSource,
+				generateSql(properties.getTableName(),columnExpressionVariations.keySet()));
+		jdbcMessageHandler.setSqlParameterSourceFactory(
```
ericbottard, line 97: Wondering if/how `ExpressionEvaluatingSqlParameterSourceFactory` could be used here.

ericbottard:
Why drop the batching support? Could have our own, at least till SI provides what we need, WDYT?

trisberg:
I thought we decided to wait for it to be in SI -- https://jira.spring.io/browse/INT-3530

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg, line 97: Don't think it supports multiple expressions for a column

ericbottard, line 97: Could we fake it by constructing a single expression using the elvis operator?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/DefaultInitializationScriptResource.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.springframework.core.io.ByteArrayResource;
+
+/**
+ * An in-memory script crafted for dropping-creating the table we're working with.
+ * All columns are created as VARCHAR(2000).
+ *
+ * @author Eric Bottard
+ */
+public class DefaultInitializationScriptResource extends ByteArrayResource {
```
artembilan, line 32: Any arguments that this class should be `public`, but not an inner one in the `JdbcSinkConfiguration` ?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/DefaultInitializationScriptResource.java
```diff
@@ -0,0 +1,56 @@
...
+		StringBuilder result = new StringBuilder("DROP TABLE ");
+		result.append(properties.getTableName()).append(";\n\n");
+
+		result.append("CREATE TABLE ").append(properties.getTableName()).append('(');
+		int i = 0;
+		for (String column : properties.getColumns().keySet()) {
+			if (i++ > 0) {
+				result.append(", ");
+			}
+			result.append(column).append(" VARCHAR(2000)");
```
artembilan, line 50: Does `VARCHAR` work for any RDBMS?
And why `2000`, but not `4000`?..

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg, line 97: Added some code to handle JsonPropertyAccessor.ToStringFriendlyJsonNode so I think we are stuck with our own implementation for now.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkApplication.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * A main application that can be used to run the Jdbc Sink as a standalone app.
+ *
+ * @author Eric Bottard
+ */
+@SpringBootApplication
+public class JdbcSinkApplication {
+
+	public static void main(String[] args) throws InterruptedException {
```
artembilan: The `throws` looks like overhead here.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+@EnableBinding(Sink.class)
+@EnableConfigurationProperties(JdbcSinkProperties.class)
+public class JdbcSinkConfiguration {
+
+	private static final Logger logger = LoggerFactory.getLogger(JdbcSinkConfiguration.class);
+
+	public static final Object NOT_SET = new Object();
+
+	@Autowired(required = false)
+	private SpelExpressionParser spelExpressionParser = new SpelExpressionParser();
```
artembilan, line 70: Interesting construction...
So, `@Autowired` will overwrite this `new` on the injection phase ?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/DefaultInitializationScriptResource.java
ericbottard, line 50: This will be enhanced in a subsequent PR. This mimics the current XD behavior

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard, line 70: Yeah, I don't think we need the autowire here. This was added when I thought we needed SI bits, but those are actually carried by the `EvaluationContext`, not the parser itself.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkApplication.java
```diff
@@ -0,0 +1,18 @@
+package org.springframework.cloud.stream.module.jdbc;
```
artembilan, line 17: No `Copyright`.
Is that intentional?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+							}
+						}
+						return parameterSource;
+					}
+				});
+		return jdbcMessageHandler;
+	}
+
+	@ServiceActivator(autoStartup = "false", inputChannel = Sink.INPUT)
+	public void handleMessage(Message<?> msg) {
```
artembilan: This isn't only redundant, but also some overhead.
The `@ServiceActivator` declaration should be moved to the `jdbcMessageHandler` `@Bean` definition.
The same concern which I raised in the Cassandra Sink PR before.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+	@Bean
+	public DataSourceInitializer nonBootDataSourceInitializer(DataSource dataSource, ResourceLoader resourceLoader) {
+		DataSourceInitializer dataSourceInitializer = new DataSourceInitializer();
+		dataSourceInitializer.setDataSource(dataSource);
+		ResourceDatabasePopulator databasePopulator = new ResourceDatabasePopulator();
+		databasePopulator.setIgnoreFailedDrops(true);
+		dataSourceInitializer.setDatabasePopulator(databasePopulator);
+		if ("true".equals(properties.getInitialize())) {
+			databasePopulator.addScript(new DefaultInitializationScriptResource(properties));
+		} else {
```
artembilan: Minor, of course, but anyway `else` on new line.
Some IDE code formatting settings vulnerabilities.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+	@ConditionalOnProperty("initialize")
+	@Bean
+	public DataSourceInitializer nonBootDataSourceInitializer(DataSource dataSource, ResourceLoader resourceLoader) {
+		DataSourceInitializer dataSourceInitializer = new DataSourceInitializer();
+		dataSourceInitializer.setDataSource(dataSource);
+		ResourceDatabasePopulator databasePopulator = new ResourceDatabasePopulator();
+		databasePopulator.setIgnoreFailedDrops(true);
+		dataSourceInitializer.setDatabasePopulator(databasePopulator);
+		if ("true".equals(properties.getInitialize())) {
+			databasePopulator.addScript(new DefaultInitializationScriptResource(properties));
```
artembilan, line 165: Looking at this one more time, I really don't like this `DefaultInitializationScriptResource`.
You should just build an appropriate String from the properties and use its `getBytes()` for the regular `ByteArrayResource`.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+		} else {
+			databasePopulator.addScript(resourceLoader.getResource(properties.getInitialize()));
+		}
+		return dataSourceInitializer;
+	}
+
+	@Bean
+	public JdbcOperations jdbcOperations(DataSource dataSource) {
```
artembilan: This is redundant bean

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+						Message<?> message = (Message<?>) o;
+						MapSqlParameterSource parameterSource = new MapSqlParameterSource();
+						for (String key: columnExpressionVariations.keySet()) {
+							List<Expression> spels = columnExpressionVariations.get(key);
+							Object value = NOT_SET;
+							EvaluationException lastException = null;
+							for (Expression spel : spels) {
+								try {
+									value = spel.getValue(evaluationContext, message);
+								}
```
artembilan, line 114: Looks like you have missed `break` here if the first expression evaluates correctly.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+		@Bean
+		@ConfigurationPropertiesBinding
+		public ShorthandMapConverter shorthandMapConverter() {
+			return new ShorthandMapConverter();
+		}
+	}
+
+	@PostConstruct
+	public void afterPropertiesSet() {
+		this.evaluationContext = IntegrationContextUtils.getEvaluationContext(beanFactory);
```
artembilan: Should be changed to the `ExpressionUtils.createStandardEvaluationContext(beanFactory)` - creates `StandardEvaluationContext` if there is no one in the ctx.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+		return new JdbcTemplate(dataSource);
+	}
+
+	/*
+	 * This is needed to prevent a circular dependency issue with the creation of the converter.
+	 */
+	public static class Nested {
+		@Bean
+		@ConfigurationPropertiesBinding
+		public ShorthandMapConverter shorthandMapConverter() {
```
artembilan, line 179: I wonder how is it resolved if we don't have here `@Configuration` or even `@Component` for `light` `@Bean`?..

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverter.java
```diff
@@ -0,0 +1,66 @@
...
+		String[] mappings = source.split("(?<!\\\\),");
+		for (String mapping : mappings) {
+			// Turn backslash-comma back to comma
+			String unescaped = mapping.trim().replace("\\,", ",");
+			if (unescaped.length() == 0) {
+				continue;
+			}
+			// Split on colon, if not preceded by backslash
+			String[] keyValuePair = unescaped.split("(?<!\\\\):");
+			Assert.isTrue(keyValuePair.length <= 2, "'" + unescaped + "' could not be parsed to a 'key:value' pair or simple 'key' with implicit value");
```
artembilan: Very long line of code.
Minor,  of course, but I wonder if we follow with some code style rules...

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverter.java
```diff
@@ -0,0 +1,66 @@
...
+ * A Converter from String to Map that accepts csv {@literal key:value} pairs
+ * (similar to what comes out of the box in Spring Core) but also simple
+ * {@literal key} items, in which case the value is assumed to be equal to the key.
+ *
+ * <p>Additionally, commas and colons can be escaped by using a backslash, which is
+ * useful if said mappings are to be used for SpEL for example.</p>
+ *
+ * @author Eric Bottard
+ */
+public class ShorthandMapConverter implements Converter<String, Map<String, String>>, ConditionalConverter{
```
artembilan, line 37: Seems for me this is a good candidate to be moved somewhere into `common`

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/SupportsShorthands.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * This annotation, when set on a Map field or setter, instructs the conversion service
+ * to use the special {@link ShorthandMapConverter}.
+ *
+ * @author Eric Bottard
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD, ElementType.METHOD})
+public @interface SupportsShorthands {
```
artembilan, line 32: Why can't we bypass this one?

jdbc-sink/src/test/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverterTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.jdbc;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.hamcrest.collection.IsMapContaining.*;
+import static org.junit.Assert.*;
```
artembilan: Is it OK to use asterisk imports?

jdbc-sink/src/test/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverterTest.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+
+import org.junit.Test;
+
+/**
+ * Unit tests for ShorthandMapConverter.
+ *
+ * @author Eric Bottard
+ */
+public class ShorthandMapConverterTest {
```
artembilan: Must be in plural. Isn't it ?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+	private JdbcMessageHandler msgHandler;
+
+
+	@Bean
+	public JdbcMessageHandler jdbcMessageHandler(DataSource dataSource) {
+		final MultiValueMap<String, Expression> columnExpressionVariations = new LinkedMultiValueMap<>();
+		for (Map.Entry<String, String> entry : properties.getColumns().entrySet()) {
+			String value = entry.getValue();
+			columnExpressionVariations.add(entry.getKey(), spelExpressionParser.parseExpression(value));
+			if (!value.startsWith("payload")) {
```
artembilan, line 91: I find this like a redundant trick.
Let's consider if I'd like to use some `header` from message or even just `new java.util.Date()`.
Right, I know that you don't (shouldn't) reach the second expression if the first one evaluates fine, but with this shorthand "hell" we can't reuse an existing `ExpressionEvaluatingSqlParameterSourceFactory`.

For the `payload` prefix I can agree for the case like `columns=a,b`

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/SupportsShorthands.java
ericbottard, line 32: What do you mean?

jdbc-sink/src/test/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverterTest.java
ericbottard: This is weird. Unless I am mistaken, I am using the sanctioned `spring-framework.xml` IDEA settings, which have static star imports set at 1:
image

Or is this IDEA playing tricks on me again?

jdbc-sink/src/test/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverterTest.java
ericbottard: correct

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverter.java
ericbottard: will split

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard: Then what's the point of `IntegrationContextUtils.getEvaluationContext`?
We want the one from SI, where we know some useful `PropertyAccessors` will have been registered

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard, line 179: You mean on the class?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard: You mean it's already added by boot?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard: +1

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard, line 91: Not sure what you mean. So, what do you suggest?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
artembilan: No, it isn't. I just mean that we don't use it at all.
Well, I see you use it from test-case. 
Do, we really need this extra (and very simple) bean in production anyway ? :smile:

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverter.java
artembilan, line 37: Pay attention, please #56

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard: Ah. Maybe it's a leftover from my previous implementation then. I'll have a look

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
artembilan: The `IntegrationContextUtils` deals only with the `BeanFactory`.
The `ExpressionUtils` delegates to the `IntegrationContextUtils` for the first attempt and fallbacks to the `StandardEvaluationContext` if there is no that one in the `BF`.
Just take a look to its source code.

I agree that it feels weird to have several similar important utility methods, but that is the historical result. We can revise it in the future release, but I'm not sure that it will simplify our life, just because we will be forced to refactor everything one more time :smile:

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
artembilan, line 179: Yes, `@Configuration` on this `Nested`  class.
Just explain for me how it works without that.
Thanks

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/SupportsShorthands.java
artembilan, line 32: I mean if we would be able to live just with the `Converter<String, Map<String, String>>` condition, without an extra annotation do not forget to add to accept the `ShorthandMapConverter` for the property.
Or do we have here the collision with `StringToPropertiesConverter`?

jdbc-sink/src/test/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverterTest.java
artembilan: I think you are correct here: https://github.com/spring-projects/spring-framework/blob/master/src/idea/spring-framework.xml#L16.

And SF team really uses asterisk imports: 
https://github.com/spring-projects/spring-framework/blob/master/spring-websocket/src/test/java/org/springframework/web/socket/client/WebSocketConnectionManagerTests.java#L37

https://github.com/spring-projects/spring-framework/blob/master/spring-jms/src/test/java/org/springframework/jms/annotation/EnableJmsTests.java#L47

But should we do the same in our projects?..

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
artembilan, line 91: I suggest for case of `payload` always request this prefix:

```
columns=a: payload.a, b
```

where `b` will be converted to the `b: payload.b`, like you do right now.
Only single `expression` and, therefore, `ExpressionEvaluatingSqlParameterSourceFactory` gain.
Just because I will be able to do this:

```
columns=a: payload.a, b: headers.b
```

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/SupportsShorthands.java
ericbottard, line 32: Not only that, but not all `Properties` injection points want the "shorthand" behavior that `ShorthandMapConverter` comes with

artembilan, line 32: OK. That's fine with me. 
So, shouldn't we move these both guys somewhere to `common` to be reusable, like that Groovy Script variables case ?

ericbottard, line 32: We could indeed. I would wait for another consumer of this specific need though, what do you think?

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkApplication.java
trisberg: More like noise, but doesn't seem necessary. Removing it.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg, line 70: Removing the autowire.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkApplication.java
trisberg, line 17: Adding it.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg: Ahh, yes, moving it.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg: Fixed.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg: Removing it.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg, line 114: Adding a break.

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
trisberg: Changed to use `ExpressionUtils`

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/ShorthandMapConverter.java
trisberg: done

artembilan:
> I thought we decided to wait for it to be in SI

Guys, what would you like to see there?
Any thought how to address? 

Thanks

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard, line 91: The whole point of the "shorthand" trick is to quickly reference a set of props to column mappings.
If we don't implicitly assume `payload.` as a prefix, then there is no point in the shorthand notation, as `payload.x` is certainly what would be used in 95% of the cases.

I suggest to merge as is, and we can revisit, for example maybe by allowing `ExpressionEvaluatingSqlParameterSourceFactory` to take several alternatives

jdbc-sink/src/main/java/org/springframework/cloud/stream/module/jdbc/JdbcSinkConfiguration.java
ericbottard, line 179: Nested classes of `@Configuration` classes are scanned automatically (I discovered that myself recently). Even more, inner classes should not be annotated with `@Configuration` if it is profile driven (I discovered that when looking at https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/CloudConfiguration.java#L45-L69

ericbottard:
Merged as 9643875 alongside https://github.com/spring-cloud/spring-cloud-dataflow/commit/16c9db1a2138ca597c8a8dbd13966e3cece26c98

PR #57, opened 2015-11-03
Remove optional=true from some module dependencies

ericbottard:
While IMHO those dependencies should ultimately be added as includes
by the deployers that need them (XD-3615), for the time being they
should really not be marked optional.
They ARE needed at runtime my modules/deployers, but are currently 
excluded because of https://github.com/spring-cloud/spring-cloud-stream/issues/168.
Even when that is fixed (alignment with boot), they still shouldn't be marked
optional.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,6 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-lattice-core</artifactId>
 			<version>${spring-cloud-lattice.version}</version>
-			<optional>true</optional>
 		</dependency>
 		<!-- Cloud connector dependencies -->
 		<!-- Lattice connector dependency to create services info from lattice -->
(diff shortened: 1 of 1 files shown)


ericbottard:
I shall add this is a prereq to https://github.com/spring-cloud/spring-cloud-dataflow/pull/134 & https://github.com/spring-cloud/spring-cloud-dataflow/pull/132

ilayaperumalg:
The `optional` flag was added for these dependencies to avoid having these dependencies transitively when the module itself is added as dependency. Please see the related discussion [here](https://github.com/spring-cloud/spring-cloud-stream-modules/pull/24#discussion-diff-37690910). Currently, these dependencies should be part of the module executable and available when the deployer resolves the module.

ericbottard:
First of all, boot does not package deps of optional deps. It so happens that the dependencies we talk about here don't have themselves dependencies that are needed (or at least they're either re-added elsewhere / the code paths we use don't need them). But this is wrong in itself in my opinion.

Secondly, I don't really understand the rationale [here](https://github.com/spring-cloud/spring-cloud-stream-modules/pull/24#discussion-diff-37690910): if we decide that a unit module needs those deps, then a composed module ought to need them (it's a module).

Put it another way: given the way deployers currently work (again, I'd prefer to revisit via XD-3615) those dependencies are NEEDED for the lattice,marathon & k8s deployers to function properly. So, whether it's a unit module or a composition thereof, they are NEEDED.

The problem is made more apparent because of https://github.com/spring-cloud/spring-cloud-stream/issues/168, but really those deps should not have been marked optional in the first place IMHO.

ilayaperumalg:
> if we decide that a unit module needs those deps, then a composed module ought to need them (it's a module).

If you compose the modules via aggregation (or direct-binding), then I believe those connector deps are automatically available. If you create a composed module using a separate POM, then I believe one would always have the SCSM parent POM as its parent which then brings the connector dependencies.

But, I agree. We need to remove the optional flag **only** to avoid [this](https://github.com/spring-cloud/spring-cloud-stream/issues/168).

ilayaperumalg:
LGTM, rebased and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/4866d047276653203a60b34dfa86fb065e71abd2

PR #58, opened 2015-11-03
XD-3534 Add HDFS Dataset sink module

### LINKED ISSUE #105: Port HDFS DataSet as s-c-s sink
As a developer, I'd like to move *hdfs-dataset* module from XD to s-c-s repo, so I can use it as sink to build streaming pipeline. 

**Acceptance:**
* *hdfs-dataset* as sink module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *hdfs-dataset* module with respective group/artifact/version detail

trisberg:
- Ported existing Spring XD hdfs-dataset sink module

This resolves #105.


hdfs-dataset-sink/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+	<description>HDFS Dataset Sink stream module</description>
+
+	<parent>
+		<groupId>org.springframework.cloud.stream.module</groupId>
+		<artifactId>spring-cloud-stream-modules</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<start-class>org.springframework.cloud.stream.module.hdfs.dataset.DatasetSinkApplication</start-class>
```
artembilan: Looks like should be this one `org.springframework.cloud.stream.module.dataset.sink`...

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
```diff
@@ -0,0 +1,64 @@
...
+		Object payload = message.getPayload();
+		if (payload instanceof Collection<?>) {
+			Collection<?> payloads = (Collection<?>) payload;
+			if (logger.isDebugEnabled()) {
+				logger.debug("Writing a collection of " + payloads.size() +
+						" POJOs of type " + payloads.toArray()[0].getClass().getName());
+			}
+			datasetOperations.write((Collection<?>) message.getPayload());
+		} else {
+			logger.warn("Expected a collection of POJOs but received " + message.getPayload().getClass().getName());
```
artembilan: Not sure that it is a big issue just have a single POJO and store it as is without any warn...
`else` on new line, BTW

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
```diff
@@ -0,0 +1,190 @@
...
+		return new DirectChannel();
+	}
+
+	@Bean
+	@ServiceActivator(inputChannel= Sink.INPUT)
+	FactoryBean<MessageHandler> aggregatorFactoryBean(MessageChannel toSink, MessageGroupStore messageGroupStore) {
+		AggregatorFactoryBean aggregatorFactoryBean = new AggregatorFactoryBean();
+		aggregatorFactoryBean.setCorrelationStrategy(
+				new ExpressionEvaluatingCorrelationStrategy("payload.getClass().name"));
+		String releaseStrategyExpression = "size() == " + properties.getBatchSize();
```
artembilan: Looks like `MessageCountReleaseStrategy` would be better for this kind of aggregator.

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
```diff
@@ -0,0 +1,87 @@
...
+
+	@Test
+	public void testWritingSomething() throws IOException, InterruptedException {
+		sink.input().send(new GenericMessage<>("Foo"));
+		sink.input().send(new GenericMessage<>("Bar"));
+	}
+
+	@After
+	public void checkFilesClosedOK() throws IOException, InterruptedException {
+		applicationContext.close();
```
artembilan: What is the reason to `close()` ctx manually if you are based on the Spring Test Framework?
Especially it will be very dangerous when you add one more `@Test`  to this class

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkIntegrationTests.java
```diff
@@ -0,0 +1,221 @@
...
+			sink.input().send(new GenericMessage<>("Foo"));
+			sink.input().send(new GenericMessage<>("Bar"));
+			sink.input().send(new GenericMessage<>("Baz"));
+		}
+
+		@After
+		public void checkFilesClosedOK() throws IOException {
+			applicationContext.close();
+			File testOutput = new File(testDir);
+			assertTrue("Dataset path created", testOutput.exists());
```
artembilan, line 95: That looks weird to `assert` in the `@After`...

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkIntegrationTests.java
```diff
@@ -0,0 +1,221 @@
...
+			"directory=${java.io.tmpdir}/dataset",
+			"namespace=pojo",
+			"allowNullValues=false",
+			"format=avro",
+			"partitionPath=year('timestamp')",
+			"batchSize=2",
+			"idleTimeout=2000"})
+	public static class PartitionedDatasetTests extends DatasetSinkIntegrationTests {
+
+		Long systemTime = 1446570266690L; // Date is 2015/11/03
```
artembilan, line 127: What is the purpose of this?
Why `System.currentTimeMillis()` isn't enough if that?

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
```diff
@@ -0,0 +1,64 @@
...
+
+	private DatasetOperations datasetOperations;
+
+	@Autowired
+	public void setDataStoreWriter(DatasetOperations datasetOperations) {
+		this.datasetOperations = datasetOperations;
+	}
+
+	@ServiceActivator(inputChannel="toSink")
+	public void datasetSink(Message<?> message) {
```
artembilan: I'd avoid this class in favor of simple `@Bean` for an inline `MessageHandler` impl.
With this `@ServiceActivator(inputChannel="toSink")`, of course.
From other side the same can be utilized with the SI Java DSL. Especially that horror around the `AggregatorFactoryBean`...

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
```diff
@@ -0,0 +1,190 @@
...
+import org.springframework.integration.store.MessageGroupStoreReaper;
+import org.springframework.integration.store.SimpleMessageStore;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.util.StringUtils;
+
+import java.util.Arrays;
+import java.util.List;
```
artembilan, line 66: What import rules do you follow?
https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style:

> The import statements are structured as follow:

```
import java.*
import javax.*
blank line
import all other imports
blank line
import org.springframework.*
blank line
import static all other imports
```

> Also, static imports should not be used in production code. They should be used in test code, especially for things like org.junit.Assert.

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
```diff
@@ -0,0 +1,190 @@
...
+		StandardEvaluationContext ctx = new StandardEvaluationContext(psb);
+		for (String expr : expressions) {
+			try {
+				Expression e = parser.parseExpression(expr);
+				psb = e.getValue(ctx, PartitionStrategy.Builder.class);
+			}
+			catch (SpelParseException spe) {
+				if (!expr.trim().endsWith(")")) {
+					throw new StoreException("Invalid partitioning expression '" + expr
+							+ "' -  did you forget the closing parenthesis?", spe);
```
artembilan, line 215: I'm not sure that it is a good idea to say that user is fool so openly. :smile:

hdfs-dataset-sink/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+		<spring-data-hadoop.version>2.3.0.RC1</spring-data-hadoop.version>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-hadoop-boot</artifactId>
+			<version>${spring-data-hadoop.version}</version>
+			<scope>provided</scope>
```
ericbottard, line 26: I think provided should be removed here

hdfs-dataset-sink/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+					<groupId>javax.servlet.jsp</groupId>
+					<artifactId>jsp-api</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-hadoop-store</artifactId>
+			<version>${spring-data-hadoop.version}</version>
+			<scope>provided</scope>
```
ericbottard, line 58: ditto

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkApplication.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.dataset.sink;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+/**
+ * Spring Boot App for starting the {@link DatasetSink} module
```
ericbottard: Missing dot at the end

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
```diff
@@ -0,0 +1,64 @@
...
+		Object payload = message.getPayload();
+		if (payload instanceof Collection<?>) {
+			Collection<?> payloads = (Collection<?>) payload;
+			if (logger.isDebugEnabled()) {
+				logger.debug("Writing a collection of " + payloads.size() +
+						" POJOs of type " + payloads.toArray()[0].getClass().getName());
```
ericbottard: We may want to avoid `toArray()` just for logging

ericbottard: And the collection could be empty...

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
```diff
@@ -0,0 +1,190 @@
...
+
+		ExpressionParser parser = new SpelExpressionParser();
+		PartitionStrategy.Builder psb = new PartitionStrategy.Builder();
+		StandardEvaluationContext ctx = new StandardEvaluationContext(psb);
+		for (String expr : expressions) {
+			try {
+				Expression e = parser.parseExpression(expr);
+				psb = e.getValue(ctx, PartitionStrategy.Builder.class);
+			}
+			catch (SpelParseException spe) {
```
ericbottard, line 212: This would rather be caught at injection time by having an `Expression[]` filed + getter/setter

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
```diff
@@ -0,0 +1,162 @@
...
+	 */
+	private String partitionPath = null;
+
+	/**
+	 * The size of the cache to be used for partition writers (10 if omitted).
+	 */
+	private int writerCacheSize = -1;
+
+	/**
+	 * Compression type name (snappy, deflate, bzip2 (avro only) or uncompressed)
```
ericbottard, line 78: Can this be made into an enum, especially if we document those as the only acceptable values.

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
```diff
@@ -0,0 +1,162 @@
...
+	 */
+	private String format = "avro";
+
+	/**
+	 * The partition path strategy to use, a list of KiteSDK partition expressions separated by a '/' symbol.
+	 */
+	private String partitionPath = null;
+
+	/**
+	 * The size of the cache to be used for partition writers (10 if omitted).
```
ericbottard, line 73: If 10 is the default, why not initialize at 10 (and remove doc about the default)?

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * Whether null property values are allowed, if set to true then schema will use UNION for each field.
+	 */
+	private boolean allowNullValues = false;
+
+	/**
+	 * The format to use, valid options are avro and parquet.
+	 */
+	private String format = "avro";
```
ericbottard, line 65: enum?

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * The URI to use to access the Hadoop FileSystem.
+	 */
+	private String fsUri;
+
+	/**
+	 * The base directory path where the files will be written in the Hadoop FileSystem.
+	 */
+	private String directory = "/tmp/hdfs-dataset-sink";
```
ericbottard, line 39: Should this depend on the "stream name"? ie `@Value("${spring.application.name:foo}")`

ericbottard, line 39: Ah, I see the following property is built that way. nvm

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * The sub-directory under the basePath where files will be written.
```
ericbottard: this refers to `basePath` but is called `directory` above, correct?

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/domain/TestPojo.java
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.dataset.domain;
+
+/**
+ * Test POJO for writing partitioned dataset
```
ericbottard: Missing dot (and `@author`)

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Thomas Risberg
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = DatasetSinkApplication.class)
+@WebIntegrationTest({"server.port:0",
+		"spring.hadoop.fsUri=file:///",
+		"directory=${java.io.tmpdir}/dataset",
```
ericbottard: Javadoc on the Properties class says this is a Path in Hadoop

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
```diff
@@ -0,0 +1,87 @@
...
+		"namespace=test",
+		"batchSize=2",
+		"idleTimeout=2000"})
+@DirtiesContext
+public class DatasetSinkConfigurationTests {
```
ericbottard: This does not really seem to test _configuration_.
Moreover, the test about file being ok when context closes is also in another class, unless I miss something

hdfs-dataset-sink/src/test/resources/logback.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ericbottard, line 1: This should be removed

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
```diff
@@ -0,0 +1,64 @@
...
+	public void setDataStoreWriter(DatasetOperations datasetOperations) {
+		this.datasetOperations = datasetOperations;
+	}
+
+	@ServiceActivator(inputChannel="toSink")
+	public void datasetSink(Message<?> message) {
+		Object payload = message.getPayload();
+		if (payload instanceof Collection<?>) {
+			Collection<?> payloads = (Collection<?>) payload;
+			if (logger.isDebugEnabled()) {
```
jvalkeal: Others have nitpicked to me that with slf4j `logger.isDebugEnabled()` is not needed anymore if you use parameterized messages within a log statement. Thought if you don't change debug call then this is needed. Nitpicking... :)

hdfs-dataset-sink/pom.xml
trisberg: Indeed, forgot to change this after changing the package name. Fixing.

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
trisberg: It indicates something wrong in the configuration and it would be bad since writing each object individually would create lots of small files in HDFS.

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
trisberg: Just porting what we have in XD, will give `MessageCountReleaseStrategy` a try.

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
trisberg: Seems to be needed for at least one of the tests, I'll remove the ones not needed.

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkIntegrationTests.java
trisberg, line 95: Yeah, not necessary for these tests, will remove.

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkIntegrationTests.java
trisberg, line 127: Just to have some control, want the messages to always fall within the same year.

hdfs-dataset-sink/pom.xml
trisberg, line 26: We need that if we want to switch in another hadoop distro using --includes

hdfs-dataset-sink/pom.xml
trisberg, line 58: ditto

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
trisberg: changing the logging here, don't need to be this detailed

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
trisberg, line 73: if it's not specified we don't want to set this since that adds a runtime system property just for this rarely used setting, so we need to keep track of when this property is not set - all this gets handled by Kite SDK where the current default they use is 10

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
trisberg, line 65: didn't have one in XD so out of scope ;)

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
trisberg: yes, will clarify the javadoc

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
trisberg: it is, but with the fsUri set to file:/// it actually writes to the local file system through hdfs, so using a tmp directory made sense

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
trisberg: yeah, removing this test since other tests cover this

hdfs-dataset-sink/src/test/resources/logback.xml
trisberg, line 1: why, without it we get a ton of debug logging when the test class is loaded

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSink.java
trisberg: ok, changing the logging

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfiguration.java
trisberg, line 212: trying to keep the functionality from XD where partition expressions are separated by '/' character

jvalkeal:
I have some issues with `idleTimeout` and `fsUri`:

Running local `hdfs://localhost:8020` and I simply build the module locally and registered module with:

```
module register --coordinates org.springframework.cloud.stream.module:hdfs-dataset-sink:jar:exec:1.0.0.BUILD-SNAPSHOT --name hdfs-dataset-sink --type sink
```

`idleTimeout` defaults to `0` so it seems that every single write creates a new dataset so I'm getting new avro file every 2 seconds. Being honest I don't remember if XD dataset sink behaves like this. I'd expect `idleTimeout=0` to disable flushing and thus simply falling back to `batchSize` when new avro file is created.

```
stream create --name "ticktock" --definition "time --fixedDelay=2|hdfs-dataset" --deploy
```

```
drwxr-xr-x   - root supergroup          0 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data
drwxr-xr-x   - root supergroup          0 2015-11-05 07:43 /tmp/hdfs-dataset-sink/data/string
drwxr-xr-x   - root supergroup          0 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/.metadata
-rw-r--r--   3 root supergroup        173 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/.metadata/descriptor.properties
-rw-r--r--   3 root supergroup          8 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/.metadata/schema.avsc
drwxr-xr-x   - root supergroup          0 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/.metadata/schemas
-rw-r--r--   3 root supergroup          8 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/.metadata/schemas/1.avsc
-rw-r--r--   3 root supergroup        105 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/0142625d-86f4-4034-be0a-a223ca78fdba.avro
-rw-r--r--   3 root supergroup        105 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/040ce3e4-46c0-4602-a8a7-f8c2f48fb6ea.avro
-rw-r--r--   3 root supergroup        105 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/041535af-c176-438f-8744-ad53c15a9c00.avro
-rw-r--r--   3 root supergroup        116 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/1c38c57d-ac86-4140-819d-0f648f9dc84c.avro
-rw-r--r--   3 root supergroup        105 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/294c5fa0-ebb5-403d-96aa-2a76cf0f01a5.avro
-rw-r--r--   3 root supergroup        105 2015-11-05 07:42 /tmp/hdfs-dataset-sink/data/string/2c0dbc08-e98e-4286-bc37-67d67017e038.avro
```

Tried to use `20sec` `idleTimeout` but then it waits default `batchSize` `10000` so didn't get any writes. Expected and didn't want to wait batch to reach its size.

```
stream create --name "ticktock" --definition "time --fixedDelay=2|hdfs-dataset --idleTimeout=20000" --deploy
```

With `batchSize=10` I get 3 avro files per minute which is expected if `idleTimeout` is more that 2 sec.

```
stream create --name "ticktock" --definition "time --fixedDelay=2|hdfs-dataset --idleTimeout=20000 --batchSize=10" --deploy
```

I don't see `fsUri` used anywhere in `DatasetSinkConfiguration`. It seem to use auto-config for `Configuration` which defaults to `hdfs://localhost:8020`. So expected stream writes to fail if giving spoofed `fsUri`. Files were still written under custom location.

```
stream create --name "ticktock" --definition "time --fixedDelay=2|hdfs-dataset --fsUri=hdfs://localhost:12345 --directory=/tmp/hdfs-dataset-sink-custom" --deploy
```

```
root@neo:/usr/local/hadoops# hadoop/bin/hdfs dfs -ls /tmp
Found 2 items
drwxrwxrwx   - root supergroup          0 2015-09-02 08:36 /tmp/hadoop-yarn
drwxr-xr-x   - root supergroup          0 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom
root@neo:/usr/local/hadoops# hadoop/bin/hdfs dfs -ls -R /tmp/hdfs-dataset-sink-custom
drwxr-xr-x   - root supergroup          0 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data
drwxr-xr-x   - root supergroup          0 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string
drwxr-xr-x   - root supergroup          0 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/.metadata
-rw-r--r--   3 root supergroup        180 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/.metadata/descriptor.properties
-rw-r--r--   3 root supergroup          8 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/.metadata/schema.avsc
drwxr-xr-x   - root supergroup          0 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/.metadata/schemas
-rw-r--r--   3 root supergroup          8 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/.metadata/schemas/1.avsc
-rw-r--r--   3 root supergroup        110 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/0274cc54-572b-4809-b11f-ab8e18ceafa6.avro
-rw-r--r--   3 root supergroup        105 2015-11-05 08:11 /tmp/hdfs-dataset-sink-custom/data/string/033fc459-21ff-4763-84a0-2b48f8e27980.avro
```

hdfs-dataset-sink/pom.xml
ericbottard, line 26: Hmm, not sure I understand exactly how distro switching would work in your opinion.
Do you expect to create BOMs that transitively include `spring-data-hadoop-boot` and `spring-data-hadoop-store` for distros? If so, and several distros depend on different versions of the above, then the code for this module may fail at runtime anyway (given we would provide a different version of say, `org.springframework.data.hadoop.store.dataset.DatasetDefinition` from what was used at compile time)

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
ericbottard, line 73: My point is, especially if Kite is responsible for the default, text documentation of defaults can become stale. actual values that are always set don't lie

hdfs-dataset-sink/src/main/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkProperties.java
ericbottard, line 65: myyyeaahhhh, but we have better DSL code completion with enums now :)
Let's open a ticket to track those

hdfs-dataset-sink/src/test/java/org/springframework/cloud/stream/module/dataset/sink/DatasetSinkConfigurationTests.java
ericbottard: I see.
Wouldn't it be better to move the ${java.io.tmpdir} part to
`spring.hadoop.fsUri`
though (`spring.hadoop.fsUri=file://${java.io.tmpdir}`) and drop it from
`directory` ?

jvalkeal:
Yes, I'm seeing same `java.io.IOException: Filesystem closed` error. Trying to figure out where it's coming from, in case it's not caused by shutdown hook.

jvalkeal:
Interesting, it seems that I don't get `Filesystem closed` error if I do:

```
stream create --name "ticktock" --definition "time|hdfs-dataset --batchSize=20" --deploy
```

and wait first file to be written.

If I do:

```
stream create --name "ticktock" --definition "time|hdfs-dataset" --deploy
```

there's no files written(if I don't wait full default batch size) and I get error.

It's worth to check if we get this error only if no data is flushed, thus no interaction with kite, thus there would not be any existing open hadoop FileSystem's.

jvalkeal:
damn, I take it back. This error doesn't happen every stream undeploy.

jvalkeal:
Right, it has to be `FileSystem` autoclose which is i.e. called from shutdown hook. I don't see errors if I disable autoclose with:

```
stream create --name "ticktock" --definition "time|hdfs-dataset --spring.hadoop.config.fs.automatic.close=false" --deploy
```

trisberg:
Rebased

jvalkeal:
squashed and merged per https://github.com/spring-cloud/spring-cloud-stream-modules/commit/3181e73a8e3e737844e9f249bcdd76043cf03eba

PR #60, opened 2015-11-10
httpclient processor

### LINKED ISSUE #82: Port HTTP-Client as s-c-s processor
As a developer, I'd like to port *http-client* module from XD to s-c-s repo, so I can use it as *processor* module to build streaming pipeline.
**Acceptance:**
* *http-client* as processor module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *http-client* module with respective group/artifact/version detail

ericbottard:
This builds on top (and replaces) https://github.com/spring-cloud/spring-cloud-stream-modules/pull/52

This also resolves #82.


httpclient-processor/README.adoc
```diff
@@ -0,0 +1,34 @@
...
+$> java -jar target/httpclient-processor-${version}-exec.jar
+```
+
+== Module Arguments
+
+The following arguments are supported:
+
+url:: The URL to issue an http request to, as a static value.
+urlExpression:: A SpEL expression against incoming message to determine the URL to use.
+httpMethod::The kind of http method to use.
```
markfisher: missing a space after `::`

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+
+	private static final Logger LOG = LoggerFactory.getLogger(HttpClientProcessor.class);
+
+	@Autowired
+	private HttpClientProcessorProperties properties;
+
+	@Autowired
+	private RestTemplate restTemplate;
+
+	@ServiceActivator(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT, autoStartup = "false")
```
markfisher: I don't think we need `autoStartup=false` here after all, considering this is not polling but only subscribing to the input channel. Also more generally, I believe changes were made to better manage the lifecycle of components in relation to the binder lifecycle. I'll cc @mbogoevici to confirm or deny that.

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
```diff
@@ -0,0 +1,160 @@
...
+	public void setHttpMethod(HttpMethod httpMethod) {
+		this.httpMethod = httpMethod;
+	}
+
+	public Object getBody() {
+		return body;
+	}
+
+	public void setBody(Object body) {
+		if (bodyExpression != null &&
```
markfisher, line 110: This is subjective, but I think it might be clearer if we either shift the check for `null || default` expression into the validation method OR only return the default value from `getBodyExpression` if both the `bodyExpresison` and `body` fields are null.

ericbottard, line 110: The second option is problematic as the default would not be documented by the generated metadata. Your first suggestion should work though, and I tend to agree with you

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
```diff
@@ -0,0 +1,160 @@
...
+	public Class<?> getExpectedResponseType() {
+		return expectedResponseType;
+	}
+
+	public void setExpectedResponseType(Class<?> expectedResponseType) {
+		this.expectedResponseType = expectedResponseType;
+	}
+
+	@NotNull
+	public Expression getResultExpression() {
```
markfisher: At first I was thinking this should this be renamed `responseExpression`, but that might not be clear that it is an expression operating on the response. That said, it could be viewed as the expression to generate a "result" which I guess does make sense... my initial reaction was to avoid more than one term for the same thing, but perhaps it's not the same thing after all. Leaving this comment anyways, in case it leads to another idea.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
```diff
@@ -0,0 +1,179 @@
...
+		public void testRequest() {
+			channels.input().send(new GenericMessage<Object>("hi"));
+			assertThat(messageCollector.forChannel(channels.output()),
+					receivesPayloadThat(is("lo hi")));
+		}
+
+	}
+
+	@RestController
+	public static class AdditionalController {
```
markfisher, line 164: very slick! ;)

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
ericbottard: Had the exact same train of thoughts. But "response" is used in the _http response_ context, so I changed to `result`

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Tests for Http Client Processor.
+ *
+ * @author Eric Bottard
+ * @author Waldemar Hummer
+ * @author Mark Fisher
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {HttpClientProcessorApplication.class, HttpClientProcessorTests.AdditionalController.class})
+@IntegrationTest("server.port=9494") // Need to use a hardcoded port to be able to reference it below
```
ericbottard: not slick at all :(

injection of `${local.server.port}` in `@WebIntegrationTest` does not work because of a lifecycle issue...

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
markfisher: Yes, that does make sense. I think the only other word to consider - and one that is more commonly used in Spring Integration components - would be "reply".

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
markfisher: Is it worth raising an issue or is it an unchangeable lifecycle issue?

ericbottard: I think there is nothing that can be done about it, as the tomcat server needs the context to start (and hence know the `@IntegrationTest` values) to know its port

markfisher: Any possibility for using `System.setProperty`, e.g. from a `@BeforeClass` - or is that still too late?

ericbottard: Haven't tried that. My guess would be it doesn't work, as `@Value("${local.server.port}")` / the `Environment` would not be ready then, but I may be wrong

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * A processor module that makes requests to an HTTP resource and emits the 
+ * response body as a message payload. This processor can be combined, e.g., 
+ * with a time source module to periodically poll results from a HTTP resource.
+ *
+ * @author Waldemar Hummer
+ * @author Mark Fisher
+ */
+@MessageEndpoint
+public class HttpClientProcessor {
```
artembilan, line 46: I wonder what is the problem do not use Spring Integration out of the box `HttpRequestExecutingMessageHandler`: http://docs.spring.io/spring-integration/docs/latest-ga/reference/html/http.html#http-outbound ...

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+	private static final Logger LOG = LoggerFactory.getLogger(HttpClientProcessor.class);
+
+	@Autowired
+	private HttpClientProcessorProperties properties;
+
+	@Autowired
+	private RestTemplate restTemplate;
+
+	@ServiceActivator(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT, autoStartup = "false")
+	public Object transform(Message<?> message) {
```
artembilan: If it is processor, so it should be `process()`, but not `transform()`.
Nit pick, of course, but I wonder if you follow with the "best code" trend.

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
markfisher: @garyrussell confirmed this - so we can remove the 'autoStartup' attribute

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Mark Fisher
+ */
+@Configuration
+@EnableBinding(Processor.class)
+@Import(SpelExpressionConverterConfiguration.class)
+@EnableConfigurationProperties(HttpClientProcessorProperties.class)
+public class HttpClientProcessorConfiguration {
+
+	@Bean
+	public RestTemplate restTemplate() {
```
artembilan, line 39: The `RestTemplate` is Spring-"free" class. Don't see reason to overhead context with redundant bean.
From other side it would be better to allow people to configure this object. E.g. `ClientHttpRequestFactory`.
We with @garyrussell recently investigated the customer request when he want in addition to the `error-code` from server accepts its custom message as well. So, we found that only `HttpComponentsClientHttpRequestFactory` helps there. And don't forget about `HttpMessageConverter` feature. Plus `ClientHttpRequestInterceptor` and of course, SSL is the most popular.

So, I'd like to hear if you have some solution to inject custom `@Configuration` to an existing module. E.g. with some marker annotation to scan classpath for additional beans. 
With that I can agree on this `RestTemplate` bean, but with `@ConditionalOnMissingBean`, when with the custom `@Configuration` end-user can inject into the module any `RestTemplate`, e.g. `OAuth2RestTemplate` and so on.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
```diff
@@ -0,0 +1,179 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.httpclient;
+
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+import static org.springframework.cloud.stream.test.matcher.MessageQueueMatcher.*;
```
artembilan, line 21: No asterisk imports.
`static` imports in the end.
https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style.
Even if SF guys uses asterisks, we found it a bit ambiguity when we use several teststing frameworks who has the same statics methods.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	@WebIntegrationTest(value = "urlExpression='http://localhost:9494/'+payload")
+	public static class TestRequestGETWithUrlExpression extends HttpClientProcessorTests {
+
+		@Test
+		public void testRequest() {
+			channels.input().send(new GenericMessage<Object>("health"));
+			assertThat(messageCollector.forChannel(channels.output()), receivesPayloadThat(containsString("status")));
+		}
```
artembilan, line 78: This both `health` tests fail for me:

> org.springframework.web.client.HttpServerErrorException: 503 Service Unavailable

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
ericbottard, line 21: Can someone point me to the IDEA configuration file to use then cc @garyrussell ?

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
artembilan, line 78: I know why: I don't have launched Redis:

> org.springframework.data.redis.RedisConnectionFailureException: Cannot get Jedis connection; nested exception is redis.clients.jedis.exceptions.JedisConnectionException: Could not get a resource from the pool

 Hence `﻿﻿﻿RedisHealthIndicator` returns `DOWN` for me and for the whole `/health` state.

I wonder if we can use some more lightweight actuator resource.

markfisher, line 78: perhaps just add another endpoint to the `AdditionalController` ?

ericbottard, line 78: Yes, added that `AdditionalController` after the fact, no need to rely on actuator anymore. Will change

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
garyrussell, line 21: We have always put static imports first; we have never allowed wildcard imports.

But I see the [Spring "standard"](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style) puts statics last; sigh.

ericbottard, line 21: So, is the file you guys are using available somewhere?

garyrussell, line 21: I don't use IDEA (I keep trying, but always end up back with STS).

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
artembilan: Well, I'd overcome this issue with separate client/server contexts:
https://github.com/spring-projects/spring-integration-samples/blob/master/basic/web-sockets/src/test/java/org/springframework/integration/samples/websocket/standard/ApplicationTests.java
or
https://github.com/spring-projects/spring-integration/blob/master/spring-integration-websocket/src/test/java/org/springframework/integration/websocket/outbound/WebSocketOutboundMessageHandlerTests.java#L106

ericbottard: The problem is about the fact that the port needs to appear in `@WebIntegrationTest(value=)` and `${local.server.port}` doesn't get replaced in there (too early)

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
markfisher: +1

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorConfiguration.java
markfisher, line 39: Since this is intended to be an out-of-the-box module, what we allow to be configured should be based on what we can expose via `@ConfigurationProperties` (as opposed to providing a `ClientHttpRequestFactory` bean, for example). Of course, anyone can create their own similar module, possibly using this one as a starting point.

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
markfisher, line 46: @ericbottard: I'm interested in your thoughts on this. In general I do think our modules should rely on Spring Integration as much as possible, as long as the underlying component maps well to the task at hand (I raised a similar point recently for the JDBC sink, and then we agreed to make those changes). That way, we can drive features into Spring Integration also, if the goals of a module are not fully supported.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
artembilan, line 21: I sent my IDEA config to the `s-c-s` Sluck channel.

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
artembilan: +1 for `reply`. 
Or in terms of `streams` it may sound better like `output`.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
artembilan: I found how to make it working:

```
@WebIntegrationTest(value = "urlExpression='http://localhost:' + @environment.getProperty('local.server.port') + '/' + payload")
```

it's for the `TestRequestGETWithUrlExpression`.
So, any all other tests should be changed to use `urlExpression` instead.

As well as this particular line should be changed to `@WebIntegrationTest(randomPort = true)`.

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorConfiguration.java
artembilan, line 39: Well, then I'd like to see the documentation how to implement my own module.
And even better how to reuse an existing one without code copy/pasting.
As I said here: it would be great just leave everything as is but allow end-user to inject his own `RestTemplate`. Something like this:

``` java
@SpringBootApplication(scanBasePackageClasses = HttpClientProcessorApplication.class)
public class MyHttpClientProcessorApplication {

    public static void main(String[] args) {
        SpringApplication.run(MyHttpClientProcessorApplication.class, args);
    }

       @Bean
    public RestTemplate restTemplate() { 
                 .........
        }
}
```

ericbottard:
Addressed review comments and changed tests to use random port per @artembilan 's suggestion

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessor.java
ericbottard, line 46: At first glance, seems like `HttpRequestExecutingMessageHandler` does what we want indeed.
Shall I try to refactor this PR to use it?

artembilan, line 46: I can do that for you!
I respect your brain for other more cooler features, so, it really would be better when I will write these modules stuff.
E.g. I don't like that we have merged WebSocket Sink not based on the SI: https://github.com/spring-cloud/spring-cloud-stream-modules/pull/54.
We are releasing AMQP-1.5.2 and SI-4.2.2 this week, so from the next week I can take care about this and any other if you say.
M-m-m. That is if I won't go to PTO though...

artembilan, line 46: Meanwhile we can merge this as is if it meets all your requirements already.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
artembilan, line 21: Hasn't been addressed.
I mean asterisk imports.

httpclient-processor/src/test/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorTests.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Tests for Http Client Processor.
+ *
+ * @author Eric Bottard
+ * @author Waldemar Hummer
+ * @author Mark Fisher
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {HttpClientProcessorApplication.class, HttpClientProcessorTests.AdditionalController.class})
+@WebIntegrationTest(randomPort = true) // Need to use a hardcoded port to be able to reference it below
```
artembilan, line 49: The comment is redundant already :smile:

ericbottard, line 49: I thought I removed it (but CTRL-Zed the removal somehow)

httpclient-processor/src/main/java/org/springframework/cloud/stream/module/httpclient/HttpClientProcessorProperties.java
markfisher, line 110: I made the change (first suggestion) and was about to merge. However, one advantage of your approach is that any caller would get `null` for one or the other (`body` or `bodyExpression`) whereas moving this to the validation method means the caller needs to know that one takes precedence. Of course, at this point (and likely always), there is only one caller, and it does indeed give precedence to the `body` property (meaning keeping the default expression does _work_):

```
if (properties.getBody() != null) {
    body = properties.getBody();
}
else if (properties.getBodyExpression() != null) {
    body = properties.getBodyExpression().getValue(message);
}
```

That said, it feels odd to rely on that, given that the calling code should not need to know anything about a _default_ expression value (as an indicator that it probably has not been set explicitly).

markfisher, line 110: Another option that I think I prefer is to allow both to be `null`, and for that to be the default case. Rather than have a default expression of `payload`, we can simply use the payload within `HttpClientProcessor` whenever both `body` and `bodyExpression` are `null`. That will also avoid any SpEL overhead for the default case. In terms of documentation, we would just need to include in both comments (for `body` and `bodyExpression` that if neither is provided, the `payload` will be used as the request body).

artembilan, line 110: :+1:

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/6ba9a8d78143e10c2fb54b33ae9ac4343615a3a4 followed by https://github.com/spring-cloud/spring-cloud-stream-modules/commit/318090e7d28a3837591dacfebfabfc5e02dff2ee

PR #61, opened 2015-11-10
Throughput measuring modules

pperalta:
An update of work done by @cppwfs and @ericbottard for throughput measuring modules.

### DIFF
--- a/load-generator-source/README.adoc
+++ b/load-generator-source/README.adoc
@@ -1,0 +1,25 @@
+= Spring Cloud Stream Load Generator Source
+
+This module generates messages with a payload of a set size to test performance of the infrastructure.
+
+== Requirements
+
+* Java 7 or Above
+
+== Classes
+
+LoadGeneratorSourceApplication:: the Spring Boot Main Application
+LoadGeneratorSource:: generates a set number messages with a payload of a specific size
+and sends them to channel
+
+== Build
+
+```
+$> mvn package
+```
+
+== Run
+
+```
+$> java -jar target/load-generator-source-${version}-exec.jar --spring.cloud.stream.bindings.output=<where to bind> --messageSize=<sizeOfMessage> --messageCount=<number of messages to send>
+```
--- a/load-generator-source/README.adoc
+++ b/load-generator-source/README.adoc
@@ -15,7 +15,7 @@ and sends them to channel
 == Build
 
 ```
-$> mvn package
+$> mvn -s .settings.xml package
 ```
 
 == Run
--- a/load-generator-source/src/main/java/org/springframework/cloud/stream/module/loadgenerator/LoadGeneratorSource.java
+++ b/load-generator-source/src/main/java/org/springframework/cloud/stream/module/loadgenerator/LoadGeneratorSource.java
@@ -43,10 +43,10 @@ public class LoadGeneratorSource extends AbstractEndpoint {
 	private static final Logger logger = LoggerFactory.getLogger(LoadGeneratorSource.class);
 
 	@Autowired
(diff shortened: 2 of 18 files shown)


load-generator-source/src/main/java/org/springframework/cloud/stream/module/loadgenerator/LoadGeneratorSource.java
```diff
@@ -0,0 +1,98 @@
...
+ * @author Glenn Renfro
+ */
+@EnableBinding(Source.class)
+@EnableConfigurationProperties({LoadGeneratorSourceProperties.class})
+public class LoadGeneratorSource extends AbstractEndpoint {
+
+	private static final Logger logger = LoggerFactory.getLogger(LoadGeneratorSource.class);
+
+	@Autowired
+	private volatile LoadGeneratorSourceProperties config;
```
ilayaperumalg: @pperalta what is the reasoning to declare this as `volatile`. Given the config properties would never change after the initialization I wonder why to make it `volatile`.

load-generator-source/src/main/java/org/springframework/cloud/stream/module/loadgenerator/LoadGeneratorSource.java
```diff
@@ -0,0 +1,98 @@
...
+
+	@Autowired
+	private volatile Source channel;
```
ilayaperumalg: Same question on `volatile` here as well.

load-generator-source/README.adoc
```diff
@@ -0,0 +1,25 @@
...
+== Classes
+
+LoadGeneratorSourceApplication:: the Spring Boot Main Application
+LoadGeneratorSource:: generates a set number messages with a payload of a specific size
+and sends them to channel
+
+== Build
+
+```
+$> mvn package
```
ilayaperumalg: We need `-s .settings.xml` here.

load-generator-source/src/main/java/org/springframework/cloud/stream/module/loadgenerator/LoadGeneratorSource.java
pperalta: This is a good question! It depends on whether it is guaranteed that the same thread that performs the auto wiring is also the same thread that invokes `doStart()`. If so, then we don't need these fields to be `volatile`.

However, if a separate thread could invoke `doStart()` then we have to determine if updates to those fields will be visible to that thread.

So far I'm seeing conflicting information. According to this, it seems that we can do without `volatile` since the fields are only set once:

> Changes to fields made by one thread are guaranteed to be visible to other threads only under the following conditions:
> ...
> The first time a thread accesses a field of an object, it sees either the initial value of the field or a value since written by some other thread.
> 
> http://gee.cs.oswego.edu/dl/cpj/jmm.html

On the other hand, this initialization pattern is literally the textbook definition of unsafe publication according to _Java Concurrency in Practice_:

screen shot 2015-11-12 at 11 10 32 am
I'll investigate some more and post an update shortly.

pperalta: After some discussion with @markfisher I've decided to remove the `volatile` since the same thread that performs the auto wiring will also invoke `doStart()`.

pperalta:
Rebased and pushed updates from review comments

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 518c523c553d272b22db171a2d8ffe2a227b7b26

PR #62, opened 2015-11-12
XD-3572: Add pmml processor

ericbottard:
Requires https://github.com/spring-cloud/spring-cloud-stream/pull/199

This resolves spring-cloud/spring-cloud-stream-modules#97.

### DIFF
--- a/pmml-processor/pom.xml
+++ b/pmml-processor/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2015 the original author or authors.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~      http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>pmml-processor</artifactId>
+	<packaging>jar</packaging>
+	<name>Pmml Processor</name>
+	<description>Pmml Processor stream module</description>
+
+	<parent>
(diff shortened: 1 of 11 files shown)


pmml-processor/pom.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2015 the original author or authors.
```
artembilan, line 3: `Copyright` in `pom` ?

pmml-processor/src/main/java/org/springframework/cloud/stream/module/pmml/processor/PmmlProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A processor that evaluates a machine learning model stored in PMML format.
+ *
+ * @author Eric Bottard
+ */
+@EnableBinding(Processor.class)
+@EnableConfigurationProperties(PmmlProcessorProperties.class)
+@Import({CustomConversionServiceRegistrar.class, SpelExpressionConverterConfiguration.class})
```
artembilan, line 67: I wonder why such kind of stuff isn't moved to some `AutoConfiguration`, e.g. the same `ChannelBindingAutoConfiguration` for `spelConvert` `@Bean`.
I see that we can't register `GenericMapConverter` as a `@ConfigurationPropertiesBinding` because it isn't `Converter<S, T>` but `GenericConverter`. That's why `ConfigurationPropertiesBindingPostProcessor` can't accepts it because it is just restricted to the:

```
@Autowired(required = false)
@ConfigurationPropertiesBinding
public void setConverters(List<Converter<?, ?>> converters) {
    this.converters = converters;
}
```

What do Spring Boot Gurus think?

Looks like we really can overcome this:

```
if (this.conversionService == null) {
    this.conversionService = getOptionalBean(
        ConfigurableApplicationContext.CONVERSION_SERVICE_BEAN_NAME,
            ConversionService.class);
}
```

only with some custom `BPP` (or `ImportBeanDefinitionRegistrar`), but anyway on that `ChannelBindingAutoConfiguration`, not here on each module...

ericbottard, line 67: Yes, we've discussed with @markfisher the benefit of having at least the `SpelExpressionConverterConfiguration` being automatically registered. I think we'll do it (but not as part of this PR obviously, several modules will need to be updated anyway)

The case for GenericMapConverter is indeed more complex. Not only is it not a `Converter<S, T>` as you mention (this bit me for a couple of hours), but it requires a reference to the `ConversionService`. And as you see in Boot code, a custom conversion service is usually created.

I propose that we merge this PR and discuss with @dsyer about how we could ease the integration of complex converter scenarios in boot.

ericbottard:
Moving the GenericMapConverter here, as using the literal map trick as discussed in https://github.com/spring-cloud/spring-cloud-stream/pull/199 wouldn't work (this module needs access to the SpEL expressions in `outputs` individually to do a `setValue()`)

markpollack:
Closed a while ago under b1cb3a38c8b51ec57fa0058312c2c511638a9881

PR #63, opened 2015-11-12
Inline `Properties` from the `@IntegrationTest`

artembilan:
See https://github.com/spring-projects/spring-boot/issues/4384


ilayaperumalg:
LGTM, rebased and merged as 301f69c.

Issue #107, opened 2015-11-15
Cassandra sink requires user with "keyspace" CREATE access


sabbyanandan:
As a user, I'm using the *Cassandra* sink with a user having CREATE/ALTER/VIEW/DELETE permissions at the table level; however, the stream fails to deploy with:

`User 3d1a7520-e361-4837-b530-0a0670f2fc43 has no CREATE permission on <all keyspaces> or any of its parents`.

**Stack trace:**
```
ERR Caused by: com.datastax.driver.core.exceptions.UnauthorizedException: User 3d1a7520-e361-4837-b530-0a0670f2fc43 has no CREATE permission on <all keyspaces> or any of its parents
2015-11-12T13:37:59.87-0800 [APP/0]      ERR     at com.datastax.driver.core.Responses$Error.asException(Responses.java:101)
2015-11-12T13:37:59.87-0800 [APP/0]      ERR     at com.datastax.driver.core.DefaultResultSetFuture.onSet(DefaultResultSetFuture.java:149)
2015-11-12T13:37:59.87-0800 [APP/0]      ERR     at com.datastax.driver.core.RequestHandler.setFinalResult(RequestHandler.java:183)
```

PR #111, opened 2015-11-18
GH-107: Fix Cassandra `keyspace` auto-creation

artembilan:
Fixes https://github.com/spring-cloud/spring-cloud-stream-modules/issues/107
- Add `createKeyspace` option to the `CassandraProperties`
- Fix `CassandraSinkIntegrationTests` to use `random` port for Embedded Cassandra as a result of upgrade to the `cassandra-unit-2.1.9.2`
- Some other simple polishing

This resolves spring-cloud/spring-cloud-stream-modules#107


trisberg:
Do we need an option to not create the table as well when using an entity object and `query-type=INSERT`? Seems like it fails with exception if table already exists.

artembilan:
Hi, @trisberg !

Thank you for looking into this!
Well, I've changed my mind and introducing `schemaAction` as well.

From the beginning I thought that it does not make sense to worry about `schemaAction` if we don't specify an entities packages. Just because it works as is without `mapping`. See `CassandraOperations.insert(T entity)` implementation - it doesn't care about particular `CassandraPersistentEntity` and so on.

But right now I think it would be better to allow for end-user more control over its stuff. So, let's him to specify  `schemaAction` on his on risk! :smile:

trisberg:
I tried using `--schemaAction=recreate` but that fails when the table already exists. I was assuming that the table would be dropped and recreated. Am I wrong assuming this?

artembilan:
Hi, @trisberg !

Thank you for taking a look and any feedback.

Well, I've just tested it against an external Cassandra and have to agree with you: it doesn't work as expected.
Just because :

```
if (dropTables) {
    if (dropUnused || mappingContext.usesTable(table)) {
        admin.dropTable(cqlId(table.getName()));
    }
}
```

returns `false`. The:

```
public boolean usesTable(TableMetadata table) {
    return entitySetsByTableName.containsKey(table.getName());
}
```

has a bug , because the `TableMetadata.name` is `String`, but `entitySetsByTableName` `key` is of `CqlIdentifier`.
Let me check with the latest Spring Data Cassandra version and I'll come back to you soon .
Anyway this `schemaAction` make sense only if we specify `entity-base-packages`. Otherwise the `mappingContext` is empty and any action will be just void. 
We should document that a bit, I think.

artembilan:
So, here you are: https://jira.spring.io/browse/DATACASS-241!

Meanwhile we can try to follow with the `spring.cassandra.schema-action=RECREATE_DROP_UNUSED`, when it drops an entire schema and only after that will create tables according to the entities in the provided packages.
I'm sure no one use the `CREATE` (and similar) action in the production.
Anyway we should fix that S-D-Cassandra bug, but that doesn't impact our solution.

trisberg:
I created issue for the `scemaAction` fix - https://github.com/spring-cloud/spring-cloud-stream-modules/issues/118 

Also created issue for validation of option combinations - https://github.com/spring-cloud/spring-cloud-stream-modules/issues/119

So, will be merging this shortly

trisberg:
Squashed and merged via commit 1828f5502a6db3b3e947329098f921142e4f63ac

PR #112, opened 2015-11-18
Issue #98: Fix nameExpression and simplify configuration

### LINKED ISSUE #98: Counter sink does not accept SpEL expressions
As a user, I'm trying to use *counter* sink with *SpEL* expression, but I'm not able to use them in combination. It throws: **exactly one of 'name' and 'nameExpression' must be set**, as the error message.

ericbottard:
This resolves #98


markpollack:
I don't know if this pattern of inner static classes has been use elsewhere but both Intellij and mvnw didn't like it 

```
initializationError(org.springframework.cloud.stream.module.metrics.CounterSinkSimpleNameTests)  Time elapsed: 0 sec  <<< ERROR!
java.lang.Exception: No runnable methods
    at org.junit.runners.BlockJUnit4ClassRunner.validateInstanceMethods(BlockJUnit4ClassRunner.java:191)
    at org.junit.runners.BlockJUnit4ClassRunner.collectInitializationErrors(BlockJUnit4ClassRunner.java:128)
    at org.junit.runners.ParentRunner.validate(ParentRunner.java:416)
    at org.junit.runners.ParentRunner.<init>(ParentRunner.java:84)
```

Running each individual test in IntelliJ was ok.  I changed to make each test its own standard subclass with an abstract base class.

markpollack:
Applied changes to separate branch as master had changed in the meantime.  Merged as 6dd3c4662406da84523308705a4ad6a7e6d7c0be

Issue #103, opened 2015-11-14
Port field-value-counter as s-c-s sink


sabbyanandan:
As a developer, I'd like to port *field-value-counter* module from XD to s-c-s repo, so I can use it as *sink* module to build streaming pipeline.

**Acceptance:**
* *field-value-counter* as sink module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *field-value-counter* module with respective group/artifact/version

ilayaperumalg:
This  is addressed in https://github.com/spring-cloud/spring-cloud-stream-modules/pull/116

PR #116, opened 2015-11-26
Add field value counter module

ilayaperumalg:
- Add field value counter that uses in-memory/redis repository
- Refactor `counter` module so that common metric configuration/properties/etc., are moved to spring-cloud-stream-modules-common-configuration sub-project
- Add tests

This resolves spring-cloud/spring-cloud-stream-modules#103


ericbottard:
Some general thoughts, I'll try to also annotate the code where appropriate:

I think we should be very careful not to blindly copy/paste the Repository approach we had in XD for analytics. In particular, I think 
- we don't want operations we don't need. As a corollary, I don't think we want to implement Spring Data Repository
- we should clarify very early (and ideally consistently between all analytics -- hence also with boot if some build on top of boot) the contract with regard to create/reset/(operations on metrics that don't yet exist).
- I think we'll want to be able to
  - increment by something other than 1
  - increment by a SpEL expression
  - derive the counter name from a SpEL

ilayaperumalg:
The main idea on this PR is to port the field value counter functionality from XD and I believe we need to have some `repository` approach to tackle the REST API for the field value counter and I thought having `spring-data` dependency would make sense in some way. 

But I agree that we need to have a way to align with `spring-boot` as close as possible.

ericbottard:
I was talking about https://github.com/ilayaperumalg/spring-cloud-stream-modules/blob/field-value-counter/common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryMetricRepository.java for example. I don't think we want to inherit from SD `Repository`

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/metrics/MetricConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.expression.EvaluationContext;
+import org.springframework.integration.context.IntegrationContextUtils;
+
+/**
+ * Sets up {@link EvaluationContext} for executing SpEL expressions.
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ */
+public class MetricConfiguration {
```
ericbottard: Not sure I understand the benefit of this class. Converter injected `Expression`s, are already properly configured

ilayaperumalg: It's a common configuration class that can be shared for all the metric modules. 

>  Converter injected Expressions, are already properly configured

Do you mean they are configured even without this configuration class?

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterRepository.java
```diff
@@ -0,0 +1,55 @@
...
+	 * Decrement the FieldValueCounter for a given field name by one, creating missing counters.
+	 * 
+	 * @param name the FieldValueCounter name
+	 * @param fieldName the name of the field
+	 * @throws IllegalArgumentException in case the given name is null
+	 */
+	void decrement(String name, String fieldName);
+
+	/**
+	 * Reset the FieldValueCounter to zero for the given field name, creating missing counters.
```
ericbottard, line 50: I think we should clarify the contract here. Inexistent counter should be the same as zero, so reset should really be delete IMO

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSinkConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+ * Configuration class for the Field Value Counter.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Configuration
+@EnableConfigurationProperties(FieldValueCounterSinkProperties.class)
+public class FieldValueCounterSinkConfiguration extends MetricConfiguration {
+
+	@Bean
+	@ConditionalOnExpression("'${store}'.isEmpty()")
```
ericbottard: weird construct. Don't we want something similar to https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/LocalConfiguration.java#L38

field-value-counter-sink/src/test/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSinkTests.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = FieldValueCounterSinkApplication.class)
+@WebIntegrationTest({"server.port:0", "name:FVCounter", "store:redis", "fieldName:test"})
+@DirtiesContext
+public class FieldValueCounterSinkTests {
+
+	//TODO RedisAvailable Rule
```
ericbottard: The Rule is now available, this class should use it IMHO

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/metrics/MetricConfiguration.java
ericbottard: yes, as per https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java#L64

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterRepository.java
ilayaperumalg, line 50: Given the `reset` is just resetting the field value counter **value**, I am not sure why the `reset` should be considered as `delete`.

ericbottard, line 50: What I'm saying is that you don't need a counter to pre-exist (with 0) to be able to start to use. The counterpart should be then, that resetting does the same thing as deleting. But what that's really implying, when combined with my desire to limit the API to the strict minimum, is that we don't need reset + delete. Only one of the two

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSinkProperties.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@ConfigurationProperties
+public class FieldValueCounterSinkProperties extends MetricProperties {
+
+	private String fieldName;
+
+	public String getFieldName() {
```
ericbottard, line 32: Should have some not blank validation

ericbottard, line 32: Also, not necessarily part of this PR, but ability to store / increment multiple fields at once should be considered

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
```diff
@@ -0,0 +1,62 @@
...
+ * Memory backed implementation of FieldValueCounterRepository that uses a ConcurrentMap
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ *
+ */
+public class InMemoryFieldValueCounterRepository extends InMemoryMetricRepository<FieldValueCounter> implements
+		FieldValueCounterRepository {
+
+	@Override
```
ericbottard: This is not used / should not be used

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/RedisFieldValueCounterRepository.java
```diff
@@ -0,0 +1,132 @@
...
+		redisTemplate.setExposeConnection(true);
+		redisTemplate.afterPropertiesSet();
+	}
+
+	/*
+	 * Note: Handler implementations typically use increment() variants to save state. The save() contract
+	 * is to store the counter as a whole. Simplest approach is erase/rewrite.
+	 */
+	@Override
+	public <S extends FieldValueCounter> S save(S fieldValueCounter) {
```
ericbottard, line 56: This is not used / should not be used

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/RedisFieldValueCounterRepository.java
```diff
@@ -0,0 +1,132 @@
...
+			return null;
+		}
+	}
+
+	@Override
+	public void increment(String counterName, String fieldName) {
+		redisTemplate.boundZSetOps(getMetricKey(counterName)).incrementScore(fieldName, 1.0);
+	}
+
+	public void increment(String counterName, String fieldName, double score) {
```
ericbottard, line 79: This signature should be the one in the interface IMO

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/metrics/MetricProperties.java
```diff
@@ -70,6 +70,10 @@ public Expression getNameExpression() {
 		return spelExpressionParser.parseExpression(nameExpression);
 	}
 
+	public Expression getComputedNameExpression() {
+		return spelExpressionParser.parseExpression(nameExpression != null ? nameExpression : ("'" + name + "'"));
```
ericbottard: We usually use `LiteralExpression` for this.
Also, the `nameExpression` can be an `Expression` directly

counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/CounterSinkConfiguration.java
```diff
@@ -15,35 +15,16 @@
...
 /**
+ * Configuration class for the `counter-sink`.
+ *
  * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
  */
 @Configuration
-@EnableConfigurationProperties(CounterSinkProperties.class)
+@EnableConfigurationProperties(MetricProperties.class)
 public class CounterSinkConfiguration {
```
ericbottard: Seems like this class is empty after refactoring. Can the `@EnableConfigurationProperties` be moved to the CounterSink class itself ?

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounter.java
```diff
@@ -0,0 +1,117 @@
...
+	 * @return the name
+	 */
+	public String getName() {
+		return name;
+	}
+
+	/**
+	 * @return the fieldValueCount
+	 */
+	public Map<String, Double> getFieldValueCount() {
```
ericbottard: this should be plural IMO

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounter.java
```diff
@@ -0,0 +1,117 @@
...
+
+	//@PersistenceConstructor
+	public FieldValueCounter(String name, Map<String, Double> fieldValueCount) {
+		Assert.notNull(name);
+		Assert.notNull(fieldValueCount);
+		this.name = name;
+		this.fieldValueCount = fieldValueCount;
+	}
+
+	/**
```
ericbottard: All the javadocs on this class members add nothing but noise

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterRepository.java
```diff
@@ -0,0 +1,57 @@
...
+ * 
+ * The name is the id and should be unique.
+ * 
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ */
+public interface FieldValueCounterRepository {
+
+	/**
+	 * Increment the FieldValueCounter for a given field name by one, creating missing counters.
```
ericbottard: not by one, but by `score`

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterRepository.java
```diff
@@ -0,0 +1,57 @@
...
+	 * 
+	 * @param name the FieldValueCounter name
+	 * @param fieldName the name of the field
+	 * @param score the incremental value
+	 * @throws IllegalArgumentException in case the given name is null
+	 */
+	void increment(String name, String fieldName, double score);
+
+	/**
+	 * Decrement the FieldValueCounter for a given field name by one, creating missing counters.
```
ericbottard: ditto

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSink.java
```diff
@@ -0,0 +1,141 @@
...
+				fieldValueCounterRepository.increment(counterName, val.toString(), 1.0);
+			}
+		}
+		else {
+			fieldValueCounterRepository.increment(counterName, value.toString(), 1.0);
+		}
+	}
+
+	protected String computeMetricName(Message<?> message) {
+		return fvcSinkProperties.getComputedNameExpression().getValue(evaluationContext,
```
ericbottard: EvaluationContext should already be set thanks to `SpelExpressionConverterConfiguration`

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ *
+ */
+public class InMemoryFieldValueCounterRepository implements FieldValueCounterRepository {
+
+	private final ConcurrentMap<String, FieldValueCounter> map = new ConcurrentHashMap<>();
+
+	public FieldValueCounter save(FieldValueCounter fieldValueCounter) {
```
ericbottard: these methods can be removed (or at minima made private) and their implementation inlined in `getOrCreate()`

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounter.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Represents the data stored in a Counter for multiple field-value pairs. Operations on it are expected to increment or
+ * decrement the value associated with a specific field name. The name property is a friendly user assigned name, and
+ * should be unique.
+ * 
+ * @author Mark Pollack
+ * 
+ */
+public final class FieldValueCounter implements Metric {
```
ericbottard, line 31: We don't need this class anymore

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSink.java
ilayaperumalg: yeah, we autowire the `EvaluationContext` that is set by the `SpelExpressionConverterConfiguration` here.

ericbottard: Sorry I wasn't clear: The `SpelExpressionConverterConfiguration` already sets `Expression`s whose `EvaluationContext` is already correctly set, so you don't have to do it here

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/metrics/MetricProperties.java
```diff
@@ -66,11 +63,15 @@ public void setName(String name) {
...
 	}
 
-	public Expression getNameExpression() {
-		return spelExpressionParser.parseExpression(nameExpression);
+	public LiteralExpression getNameExpression() {
+		return nameExpression;
 	}
 
-	public void setNameExpression(String nameExpression) {
+	public LiteralExpression getComputedNameExpression() {
```
ericbottard: Can't be a LiteralExpression for the general case. Only if a literal `name` has been provided

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
```diff
@@ -0,0 +1,87 @@
...
+	private void modifyFieldValue(String name, String fieldName, double delta) {
+		FieldValueCounter counter = getOrCreate(name);
+		Map<String, Double> data = counter.getFieldValueCounts();
+		double count = data.containsKey(fieldName) ? data.get(fieldName) : 0;
+		data.put(fieldName, count + delta);
+		save(counter);
+	}
+
+	private FieldValueCounter getOrCreate(String name) {
+		synchronized (map) {
```
ericbottard: seems like synchonized is not needed here, given increment/decrement already protect

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
```diff
@@ -0,0 +1,87 @@
...
+	private final ConcurrentMap<String, FieldValueCounter> map = new ConcurrentHashMap<>();
+
+	@Override
+	public synchronized void increment(String name, String fieldName, double score) {
+		modifyFieldValue(name, fieldName, score);
+	}
+
+	@Override
+	public synchronized void decrement(String name, String fieldName, double score) {
+		modifyFieldValue(name, fieldName, score);
```
ericbottard: should be -score

field-value-counter-sink/src/test/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSinkTests.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = FieldValueCounterSinkApplication.class)
+@WebIntegrationTest({"server.port:0", "name:FVCounter", "store:redis", "fieldName:test"})
```
ericbottard: can use server.port=-1 and spare starting tomcat, since it is not used in the test

field-value-counter-sink/src/test/java/org/springframework/cloud/stream/module/metrics/FieldValueCounterSinkTests.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	@After
+	public void clear() {
+		fieldValueCounterRepository.reset("FVCounter", "Hi");
+		fieldValueCounterRepository.reset("FVCounter", "Hello");
+	}
+
+	@Test
+	public void testFieldNameIncrement() throws InterruptedException {
```
ericbottard: Should test decrement, as well as different payload types. Deeply nested payloads as well

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
markfisher: actually it seems like overkill to have increment/decrement `synchronized` globally... also considering it's a `ConcurrentMap` I think synchronization might be avoidable altogether. In any case, concurrency at the map level, should only be limited in the case where the key is not yet present (i.e. could be more optimistic).

ilayaperumalg: +1

field-value-counter-sink/src/main/java/org/springframework/cloud/stream/module/metrics/InMemoryFieldValueCounterRepository.java
ilayaperumalg: ah, sorry.

ericbottard:
Merged as d9756a03d231eed426ea2811896a11cbbdcc20de

PR #124, opened 2015-12-10
GH-123 HttpSource default mapping/port

### LINKED ISSUE #123: HttpSource: Retain default http access to '/' as opposed to '/messages'
As a user, I'd like to access the http port at the default `/` path [as supported by Spring XD], so it is backward compatible. Currently, I'm expected to include `/messages` in order to post messages to http port.

Acceptance:

- I'm able to post data to http port with default `/` path
- Accessing `/messages` path results in 404; in other words, it should be unsupported
- Related samples/examples are updated to reflect this change

ilayaperumalg:
- Change http-source default mapping to `/` and port to `9000`
- Add test

This resolves spring-cloud/spring-cloud-stream-modules#123

### DIFF
--- a/http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceApplication.java
+++ b/http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceApplication.java
@@ -16,6 +16,10 @@
 
 package org.springframework.cloud.stream.module.http;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
@@ -23,12 +27,27 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
  * A main application that can be used to run the HTTP Source as a standalone app.
  *
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
 @SpringBootApplication
 public class HttpSourceApplication {
 
+	private static final String SERVER_PORT_OPTION = "server.port";
+
 	public static void main(String[] args) {
-		SpringApplication.run(HttpSourceApplication.class, args);
+		List<String> argsList = new ArrayList<>();
+		argsList.addAll(Arrays.asList(args));
+		boolean hasServerPort = false;
+		for (String arg : argsList) {
+			if (arg.contains(SERVER_PORT_OPTION)) {
+				hasServerPort = true;
+				return;
+			}
+		}
+		if (!hasServerPort) {
+			argsList.add("--" + SERVER_PORT_OPTION + "=9000");
+		}
+		SpringApplication.run(HttpSourceApplication.class, argsList.toArray(new String[0]));
 	}
 
 }
--- a/http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceApplication.java
+++ b/http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceApplication.java
@@ -16,10 +16,6 @@
 
 package org.springframework.cloud.stream.module.http;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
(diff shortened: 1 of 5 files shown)


http-source/src/main/java/org/springframework/cloud/stream/module/http/HttpSourceApplication.java
```diff
@@ -23,12 +27,27 @@
...
+
 	public static void main(String[] args) {
-		SpringApplication.run(HttpSourceApplication.class, args);
+		List<String> argsList = new ArrayList<>();
+		argsList.addAll(Arrays.asList(args));
+		boolean hasServerPort = false;
+		for (String arg : argsList) {
+			if (arg.contains(SERVER_PORT_OPTION)) {
+				hasServerPort = true;
+				return;
```
markfisher: I don't think `return` is intentional here.

http-source/src/test/java/org/springframework/cloud/stream/module/http/HttpSourceTests.java
```diff
@@ -54,7 +54,7 @@
  */
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringApplicationConfiguration(classes = HttpSourceApplication.class)
-@WebIntegrationTest(randomPort = true)
+//@WebIntegrationTest(randomPort = true)
```
markfisher: should be deleted, not just commented-out

markfisher:
I don't understand why this PR is doing anything with the `server.port`. Now that every module is a boot-based microservice, we need consistency, even for an HTTP source. A user can always choose port 9000 if they want (and it's available), but the default is now boot's default. For that reason, in terms of our backwards compatibility, the endpoint mapping is as far as we should go (not also the port).

ilayaperumalg:
@markfisher that makes sense. I was considering the `port` as well for the backwards compatibility. will fix it.

garyrussell:
I am a bit late to the game here, but why is this source not using Spring Integration and using MVC directly? Seems like we're missing lots of features such as payload conversion, header mapping, custom headers, etc, etc.

There was even a [SO question](http://stackoverflow.com/questions/34184995/allow-cors-in-spring-xd-http-source) yesterday about CORS support in XD.

ilayaperumalg:
I believe the `http` source SCS module was developed to use boot's embedded servlet container to get the most out of spring-boot itself. We could consider adding a flavour of `http-source` that uses SI.

garyrussell:
The SI http adapter will still use Boot's embedded servlet container - it sits on top of MVC.

ilayaperumalg:
@garyrussell Created a story to use SI adapters: https://github.com/spring-cloud/spring-cloud-stream-modules/issues/125

http-source/src/test/java/org/springframework/cloud/stream/module/http/DefaultHttpSourceTests.java
```diff
@@ -0,0 +1,74 @@
...
+	@Value("${local.server.port}")
+	protected int port;
+
+	@Autowired
+	protected MessageCollector messageCollector;
+
+	protected RestTemplate restTemplate = new RestTemplate();
+
+	@WebIntegrationTest(randomPort = true)
+	public static class DefaultMappingTests extends DefaultHttpSourceTests {
```
markfisher, line 64: isn't this a duplicate of the inner class within `HttpSourceTests` ?

markfisher, line 64: This one does use `randomPort` while the other one does not, so it's probably worth keeping... but as a follow-up, let's discuss the possibility of organizing the structure of these tests more clearly. I will go ahead and merge as-is.

http-source/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -4,7 +4,7 @@
       "name": "pathPattern",
       "type": "java.lang.String",
       "description": "An Ant-Style pattern to determine which http requests will be captured",
-      "defaultValue": "/messages"
+      "defaultValue": "/"
```
artembilan, line 7: I suppose Docs should be fixed to address this change.
And, of course, it would be better to do that during this PR.
I mean something like:

```
dataflow:>http post --contentType 'application/json' --data '{"isbn": "1599869772", "title": "The Art of War", "author": "Sun Tzu"}' --target http://localhost:8888/messages
  > POST (application/json;charset=UTF-8) http://localhost:8888/messages {"isbn": "1599869772", "title": "The Art of War", "author": "Sun Tzu"}
  > 202 ACCEPTED
```

in `sinks.adoc`

markfisher, line 7: Good catch. I pushed that change in an additional commit while merging.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/2882124fc4a097b84ff2c03f85596d2677eb3f9f along with https://github.com/spring-cloud/spring-cloud-stream-modules/commit/0d2bfa515990ab20228f0e9b133ea31c0cef2a07

PR #142, opened 2015-12-20
GH-93: Add Splitter Processor

### LINKED ISSUE #93: Port Splitter as s-c-s processor
As a developer, I'd like to port *splitter* module from XD to s-c-s repo, so I can use it as *processor* module to build streaming pipeline.
**Acceptance:**
* *splitter* as processor module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *splitter* module with respective group/artifact/version detail

garyrussell:
Resolves #93

This is a bit more sophisticated than the XD version in that the
type of splitter that's used depends on the properties:
- DefaulMessageSplitter
- ExpressionEvaluatingMessageSplitter
- FileSplitter

### DIFF
--- a/splitter-processor/README.adoc
+++ b/splitter-processor/README.adoc
@@ -1,0 +1,33 @@
+= Spring Cloud Stream Splitter Processor
+
+A *Spring Cloud Stream* module that allows one to split a message payload into component parts.
+
+== Requirements
+
+* Java 7 or Above
+
+== Code Tour
+
+SplitterProcessorApplication:: the Spring Boot Main Application
+SplitterProcessor:: the bean definitions
+SplitterProcessorProperties:: defines the configuration properties that are available to the Splitter Processor
+  * expression: a SpEL expression used to split messages
+  * delimiters: delimiters to use when splitting String payloads
+  * charset: the charset used to convert to String when splitting `File` payloads with text content
+  * fileMarkers: whether or not to emit START/END marker messagees when splitting `File` payloads with text content
+  * applySequence: whether or not to add correlation/sequence information headers (default true)
+
+Mixed payload types are not supported (with the exception that the `String` and `Collection` can be split by a single
+splitter).
+
+## Building with Maven
+
+```
+$> mvn package
+```
+
+## Running the Application
+
+```
+$> java -jar target/transform-processor-${version}-exec.jar
+```
--- a/splitter-processor/README.adoc
+++ b/splitter-processor/README.adoc
@@ -29,5 +29,5 @@ $> mvn package
 ## Running the Application
 
 ```
-$> java -jar target/transform-processor-${version}-exec.jar
+$> java -jar target/splitter-processor-${version}-exec.jar
 ```
--- a/splitter-processor/src/main/java/org/springframework/cloud/stream/module/splitter/SplitterProcessorProperties.java
+++ b/splitter-processor/src/main/java/org/springframework/cloud/stream/module/splitter/SplitterProcessorProperties.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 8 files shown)


splitter-processor/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+## Building with Maven
+
+```
+$> mvn package
+```
+
+## Running the Application
+
+```
+$> java -jar target/transform-processor-${version}-exec.jar
```
artembilan: s/transform/splitter

spring-cloud-stream-modules-docs/src/main/asciidoc/processors.adoc
```diff
@@ -91,3 +91,35 @@ This transform will convert all message payloads to upper case. If sending the w
...
+
+Ambiguous properties are not allowed.
+
+==== Extract the value of a specific field
+
+This splitter converts a JSON message payload to the value of a specific JSON field.
+
+    xd:> stream create --name jsontransformtest --definition "http --port=9005 | splitter --expression=#jsonPath(payload,'$.firstName') | log" --deploy
+
+Try sending this payload to the HTTP endpoint and you should see just the value "John" in the XD log:
```
artembilan: The `XD` word is wrong here.
From other side this sample doesn't seem appropriate for the `splitter` case. Looks like it is about `transform`.
Yes, you'll see the `John` as a result, but we should have here more complicated JSON and a bit different jsonPath to demonstrate splitter. See, `JsonPathTests` in SI Core for example.

artembilan:
One more note: let me know your position here about `XPathMessageSplitter`.
Right we can reach almost the same using `#xpath()` SpEL-function, but anyway just in case.
Thank you!

spring-cloud-stream-modules-docs/src/main/asciidoc/processors.adoc
```diff
@@ -91,3 +91,35 @@ This transform will convert all message payloads to upper case. If sending the w
...
+$$delimiters$$:: $$A list of delimiters to tokenize a String payload ('expression' must be null)$$ *($$String$$, default: `null`)*
+$$fileMarkers$$:: $$Split File payloads, when true, START and END marker messages will be emitted, when false no markers are emitted$$ *($$String$$, default: `null`)*
+$$charset$$:: $$Split File payloads using this charset to convert bytes to String$$ *($$String$$, default: `null`)*
+$$applySequence$$:: $$Add correlation and sequence information to the message headers$$ *($$String$$, default: `true`)*
+
+When no `expression`, `fileMarkers`, or `charset` is provided, a `DefaultMessageSplitter` is configured with (optional) `delimiters`.
+When `fileMarkers` or `charset` is provided, a `FileSplitter` is configured.
+Otherwise, an `ExpressionEvaluatingMessageSplitter` is configured.
+
+When splitting `File` payloads, the `sequenceSize` header is zero because the size cannot be determined at the beginning.
```
artembilan, line 111: We don't determine the `Splitter` impl at runtime when it is `File` payload. So, I think we should say something here, when you'd like to split `File`, but aren't interested in the `markers` and everything is fine with the file `charset`, you should use `--fileMarkers=false` anyway to avoid fallback to the `DefaultMessageSplitter`.

garyrussell:
Pushed - I don't think it's worth adding more complication, given that we have `#xpath`.

splitter-processor/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+== Code Tour
+
+SplitterProcessorApplication:: the Spring Boot Main Application
+SplitterProcessor:: the bean definitions
+SplitterProcessorProperties:: defines the configuration properties that are available to the Splitter Processor
+  * expression: a SpEL expression used to split messages
+  * delimiters: delimiters to use when splitting String payloads
+  * charset: the charset used to convert to String when splitting `File` payloads with text content
+  * fileMarkers: whether or not to emit START/END marker messagees when splitting `File` payloads with text content
```
artembilan, line 17: Typo: message~~e~~s

Will fix on merge.

splitter-processor/src/main/java/org/springframework/cloud/stream/module/splitter/SplitterProcessorProperties.java
```diff
@@ -0,0 +1,113 @@
...
+	public void setApplySequence(boolean applySequence) {
+		this.applySequence = applySequence;
+	}
+
+	@AssertTrue(message = "'delimiters' is not allowed when an 'expression' is provided")
+	public boolean isDelimitersAllowed() {
+		return this.expression != null ? this.delimiters == null : true;
+	}
+
+	@AssertTrue(message = "File properties are not allowed when an 'expression' or 'delimeters' property is provided")
```
artembilan, line 107: Typo: delim~~e~~**i**ters

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/1ba3dbed558e8c2cac12523b1b0779e8a5177828

PR #146, opened 2015-12-24
Add trigger source

### LINKED ISSUE #87: Port Trigger as s-c-s source
As a developer, I'd like to move *trigger* module from XD to s-c-s repo, so I can use it as source to build streaming pipeline.
**Acceptance:**
* *trigger* as source module is available in *spring-cloud-stream-modules* repo
* maven repo includes *trigger* module with respective group/artifact/version detail

ilayaperumalg:
- Add trigger source module with support for `fixedDelay`, `date` and `cron` trigger options
- Have the configuration classes for all these flavors under common modules package
  - This will enable re-use of multiple trigger flavors
- Add tests

This resolves #87


common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/CronTriggerConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.scheduling.support.CronTrigger;
+
+/**
+ * Configuration class for Cron trigger.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Configuration
+@EnableConfigurationProperties(CronTriggerProperties.class)
+@ConditionalOnProperty("cron")
```
artembilan: I'd say that we can combine all these trigger options in the single `TriggerProperties` and then process them in the single `TriggerConfiguration`, thanks to the `@Conditional` nature on the `@Bean` method level.
But that is just a matter of opinion and flavor. So, we can live with this solution as well.
But I should admit that it is a bit tough to review it with this way, having only single `PollerMetadata` in the SI...

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/DateTriggerConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+	Trigger dateTrigger() {
+		Date date;
+		DateFormat dateFormat = new SimpleDateFormat(dateTriggerProperties.getDateFormat());
+		String dateString = dateTriggerProperties.getDate() != null ? dateTriggerProperties.getDate() :
+				dateFormat.format(new Date().toString());
+		try {
+				date = dateFormat.parse(dateString);
+		}
+		catch (ParseException pe) {
+			throw new RuntimeException("Exception parsing the Date format: " + dateTriggerProperties.getDateFormat(), pe);
```
artembilan: The `BeanCreationException` would be better here. Just because we can't create this bean during the application refresh phase.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/DateTriggerConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+@ConditionalOnProperty("date")
+public class DateTriggerConfiguration {
+
+	@Autowired
+	private DateTriggerProperties dateTriggerProperties;
+
+	@Bean(name = TriggerConstants.TRIGGER_BEAN_NAME)
+	Trigger dateTrigger() {
+		Date date;
+		DateFormat dateFormat = new SimpleDateFormat(dateTriggerProperties.getDateFormat());
```
artembilan: Seems for me the `* @exception IllegalArgumentException if the given pattern is invalid` isn't appropriate here and should be as validation in the `DateTriggerProperties` and this configuration should rely only on the `DateFormat` from there.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/DateTriggerConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+		String dateString = dateTriggerProperties.getDate() != null ? dateTriggerProperties.getDate() :
+				dateFormat.format(new Date().toString());
```
artembilan: Looks like overhead: `new Date().toString()` to `parse` to `Date()` on the next line...

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerConstants.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.module.trigger;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+public class TriggerConstants {
+
+	public static final String TRIGGER_BEAN_NAME = "SPRING_CLOUD_STREAM_TRIGGER";
```
artembilan: I don't like this name.
I really can run any module outside of the cloud.
Why simple `pollerTrigger` name isn't enough for you?
Moreover this definition is in the `module` package...

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.integration.annotation.InboundChannelAdapter;
+import org.springframework.integration.annotation.Poller;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@EnableBinding(Source.class)
+@EnableConfigurationProperties(TriggerSourceProperties.class)
+@Import({PeriodicTriggerConfiguration.class, CronTriggerConfiguration.class, DateTriggerConfiguration.class,
+		SpelExpressionConverterConfiguration.class})
```
artembilan: No one injected `@ConfigurationProperties` provides any `Expression` property.
It doesn't hurt, of course, but confusing.
That's why I said no one time that `SpelExpressionConverterConfiguration` should always be imported somewhere on the `@EnableBinding`.
CC @ericbottard

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,53 @@
...
+	private EvaluationContext evaluationContext;
+
+	@Autowired
+	private TriggerSourceProperties config;
+
+	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller(
+			trigger = TriggerConstants.TRIGGER_BEAN_NAME, maxMessagesPerPoll = "1"), autoStartup = "false")
+	public String trigger() {
+		if (this.config.getPayload() != null && !this.config.getPayload().isEmpty()) {
+			return new LiteralExpression(this.config.getPayload()).getValue(this.evaluationContext, String.class);
```
artembilan: Looks like `this.config.getPayload()` is some **static** value, so it looks very strange to create `LiteralExpression` for `String`, moreover on each polling task.
How about really consider that `payload` as a runtime `Expression`? 

```
--payload='new java.util.Date()'
--payload='T(java.util.UUID).randomUUID()'
```

Or as you'd like with your test-case:

```
--payload='''cronTest'''
```

?

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,53 @@
...
+		}
+		// Return empty string as payload
+		return "";
```
artembilan: Seems for me this _default_ should be a part of `TriggerSourceProperties`

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,53 @@
...
+public class TriggerSource {
+
+	@Autowired
+	private EvaluationContext evaluationContext;
+
+	@Autowired
+	private TriggerSourceProperties config;
+
+	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller(
+			trigger = TriggerConstants.TRIGGER_BEAN_NAME, maxMessagesPerPoll = "1"), autoStartup = "false")
```
artembilan: You use `TriggerSourceProperties extends MaxMessagesProperties`, but don't care about that `maxMessages`.
See `FileSource` how to build `PollerMetadata`:

```
@Bean
public PollerMetadata poller() {
    return Pollers.trigger(trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
}
```

and after that:

```
@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller("poller"), autoStartup = "false")
```

BTW, `FileSource` uses `MaxMessagesProperties` directly without `extends`, like you have here with `TriggerSourceProperties`.
Also I can agree that `maxMessages` is very tied with `TriggerSource`. So, I can live with that `extends` :smile: .

trigger-source/src/test/java/org/springframework/cloud/stream/module/trigger/TriggerSourceTests.java
```diff
@@ -0,0 +1,72 @@
...
+	@Autowired
+	protected MessageCollector messageCollector;
+
+	@WebIntegrationTest({"fixedDelay=2", "initialDelay=1", "payload=test"})
+	public static class FixedDelayTest extends TriggerSourceTests {
+
+		@Test
+		public void fixedDelayTest() throws InterruptedException {
+			Thread.sleep(2100);
+			assertEquals(1, messageCollector.forChannel(triggerSource.output()).size());
```
artembilan: Looks like these tests are very dangerous by time...
You should use `.poll(10, TimeUnit.SECONDS);` for the `forChannel()` result.
Plus not `.equals()` for the `payload`, but `assertEquals()`.
And remove all these `Thread.sleep()`, of course.
The `take()` isn't good anyway, because we may block the thread indefinitely.

artembilan:
The module doesn't contains `README` and there is need a note in the `sources.adoc`.

Anyway it is a good job!
I'd be glad to find time to contribute here, too :smile: 

Merry Christmas, BTW!

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
ericbottard: +1 I was going to propose such a feature.
Also, it would be good if the application context for trigger contained an `AtomicInteger`, allows stuff like we had in https://github.com/spring-projects/spring-xd/blob/e5334d7934b8a6886fe1ad42515011ffd27cca61/modules/source/tcp-client/config/tcp-client.xml#L34

artembilan:
I wonder how can we standardize the `Poller` usage. See `ChannelBindingAutoConfiguration` with the `PollerMetadata.DEFAULT_POLLER` and #148 . Although that Poller on the `Binding` might be need really only for the Binder stuff and shouldn't override it...

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
ilayaperumalg: Infact this was already addressed and `SpelExpressionConverterConfiguration` is available from SCS. I will remove the configuration import from here.

artembilan:
Hey, @ilayaperumalg ! 
You haven't notified me with the `Pushed` comment about your new commit.
OK. Pulling that now locally for review...

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+
+	@Bean(name = TriggerConstants.TRIGGER_BEAN_NAME)
+	@ConditionalOnProperty(TriggerConstants.DATE_TRIGGER_OPTION)
+	Trigger dateTrigger() {
+		Date date;
+		DateFormat dateFormat = new SimpleDateFormat(config.getDateFormat());
+		String dateString = config.getDate() != null ? config.getDate() :
+				new Date().toString();
+		try {
+			date = dateFormat.parse(dateString);
```
artembilan: Seems for me we talked here about something like avoid this `date -> toString -> parse -> date` logic.
Plus I'd like to see `SimpleDateFormat` as a part of the `TriggerProperties` with some validation for the:

```
 @exception IllegalArgumentException if the given pattern is invalid
```

See `SimpleDateFormat` ctor.

Let me know otherwise. I want to understand these SCS convensions anyway.

common/spring-cloud-stream-modules-common-configuration/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerProperties.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-16 the original author or authors.
```
artembilan: Something wrong here with years...
Right, new/affected class should be updated to the 2016 already :smile: 
Minor, of course, and I can fix on merge, but need do not forget that, especially when IDE keeps Copyright collapsed...

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,61 @@
...
+	private Trigger trigger;
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.trigger(trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
+	}
+
+	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller("poller"), autoStartup = "false")
+	public String trigger() {
+		return this.config.getPayload().getValue(this.evaluationContext, String.class);
```
artembilan: It still doesn't make sense. See `TriggerSourceProperties.getPayload()`. ti returns `LiteralExpression`.
Are you going to rework it to normal runtime SpEL in the separate issue?

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,61 @@
...
+	@Autowired
+	private MaxMessagesProperties maxMessagesProperties;
+
+	@Autowired
+	private Trigger trigger;
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.trigger(trigger).maxMessagesPerPoll(this.maxMessagesProperties.getMaxMessages()).get();
+	}
```
artembilan: Looks like we can move this bean definition to the `TriggerConfiguration` to be reusable from any `Source` module.
See @joshlong 's related GH issue: #148 . It won't hurt to have `Trigger`, though, because some module may provide `adviceChain` and/or `transactionSynchronizationFactory`, but this common stuff around trigger options may be pull on top.

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller("poller"), autoStartup = "false")
```
artembilan: 1. Why `autoStartup = "false"`? The `TimeSource` doesn't do that.
2. Having this kind of annotation configuration we could introduce custom annotation:

``` java
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.RUNTIME)
@InboundChannelAdapter(value = Source.OUTPUT, poller = @Poller("poller"), autoStartup = "false")
public @interface MessageSource {

}
```

artembilan: I mean for reuse in any `@InboundChannelAdapter`-aware module.

artembilan:
That's all for this round.

ilayaperumalg:
Sorry, I missed to notify when I pushed the latest. Thanks for the review.
I will follow up on the comments.

trigger-source/src/main/java/org/springframework/cloud/stream/module/trigger/TriggerSource.java
ilayaperumalg: +1 for this. But for some reason, @Poller("poller") didn't work for me. Once figured, I will try to incorporate that change.

ilayaperumalg:
@artembilan Pushed the changes.

artembilan:
Hi there!
Here is what I think on the matter: https://github.com/artembilan/spring-cloud-stream-modules/commit/8a0531b3b1be7e59f2a93724e66205e9548c850d

ilayaperumalg:
@artembilan The changes above look good to me.

artembilan:
Hello, @ilayaperumalg !

Thank you for the feedback!
So, we need agree on the annotation name (https://github.com/artembilan/spring-cloud-stream-modules/commit/8a0531b3b1be7e59f2a93724e66205e9548c850d#diff-3d0adcd7cef4ce47490a6532b1d4cb81R30) and I'll merge to go ahead with other stuff.

artembilan:
After some discussion with @mbogoevici and @sabbyanandan we have this results:

> [2:22:49 PM] Artem Bilan: So, the annotation name from the brain storm! :)
>  [2:23:27 PM] Artem Bilan: @MessageSource
>  @PollerSource
>  @TriggerSource
>  @SourceInput
> 
> [2:25:52 PM] Marius Bogoevici: @PollerSource or @PolledSource IMO
>  [2:26:10 PM] Marius Bogoevici: @MessageSource is too general, I would say
> 
> Well, +1 for the short valuable name:
> 
> @PolledSource
> 
> Everything in those two words:
> It identify the Source module
> It says that there is some poller around 
> And it isn't Message-Driven Source like HTTP, Kafka, JMS or RabbitMQ
> 
> [2:33:51 PM] Marius Bogoevici: (y)
> 
> [2:37:28 PM] Sabby Anandan: thank you, both! (bow)

So, we finished with the `@PolledSource`, if there is any objections, there is somewhere like one hour to share your mind. Otherwise it will be merged. :smile: 

Thank you all for attention!

markfisher:
My only concern would be if "Polled" might be confused as past-tense. Would `@PollingSource` or `@PollableSource` work? (the latter probably not, as it implies something must poll _it_ as opposed to it doing the polling).

artembilan:
> [2:27:45 PM] Marius Bogoevici: No, because it’s not the source doing the polling
> [2:28:18 PM] Marius Bogoevici: the source is ‘polled’
> [2:28:25 PM] Sabby Anandan: ah!
> [2:28:28 PM] Marius Bogoevici: in the same way a dependency is ‘autowired'

mbogoevici:
My objection to `PollingSource` was that the annotated component doesn't do any polling, it is instead `polled` by the default poller. I was thinking of `polled` as an adjective, just as `autowired`.

markfisher:
but _autowired_ happens once, while polling is continuous... so maybe "pollable" does fit?

artembilan:
Yeah... That's why the general annotation is `@InboundChannelAdapter`, meaning create a `SourcePollingChannelAdapter` endpoint for this `MessageSource`...
From the SCS perspective it is really a "source of data for the stream by some polling options"...

We have `PollableChannel` abstraction for similar reason, so it might be really good to be consistent?
I mean @markfisher 's `@PollableSource` :smile:

mbogoevici:
I don't read `Polled` as a one-off thing, and think the notion of polling already implies some kind of repetition. Also `PollableChannel` is an interface, as generally '-able' things are.

That being said, I'm not necessarily opposed to `pollable`.

artembilan:
Let's end up with the `@PollingSource` as it reflects the `Source` as a module and its `Polling` behavior!
Right, we don't need (yet) any annotations for other kind of sources, but they may sound like `Listening`, `Reactive`, `Reading` etc.

sabbyanandan:
I'd still vote for `@PollingSource`, as it clearly defines the polling, which is continuous and for source modules. @mbogoevici / @markfisher: any other thoughts / concerns?

mbogoevici:
I stand by my opinion that `Polling` is not the right choice, since the annotated component doesn't actually do any polling. `Pollable` would be better, if `Polled` is not acceptable. But I don't think it's a blocker.

artembilan:
Merging with `@PollingSource`...

garyrussell:
I agree with @mbogoevici `@PollingSource` is incorrect - I prefer `@PolledSource`.

garyrussell:
Well, now I read Mark's comment, I am not so sure.

This was my rationale...

In the context of s-c-s-m the source is a noun and it is polled for messages - the fact that the polling infrastructure is contained in the source module is irrelevant

In the context of SI, SPCA properly uses “polling” because it (the channel adapter) polls the MessageSource

SourcePolledChannelAdapter would be wrong

Given that the polling is done within the source (as opposed to SI where the SPCA does the polling), polled is probably wrong.

Also, polling might be considered wrong - a Trigger might only fire once.

`@TriggeringSource` ??

artembilan:
That's why one of my suggestions was `@SourceInput` when it implies the marker for method which is really "the source of data for the stream". And it doesn't matter how it is implemented under neat.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/abc4ec5c7763e0290dd9a15dd2f05fe5e4b2c79d.

Congratulations for all of us! :smile: 

The discussion regarding `@PollableSource` name be extended anyway :wink:

PR #147, opened 2015-12-30
Increases integration  test coverage for Filter Processor and Transform Processor SCS Module

mross1080:
Added some more coverage to the Filter processor integration tests, and also introduced using the not(T matcher) so there can be negative assertions.  I didn't find any use of not(T matcher) in the rest of SCS, therefore all integration tests so far only had positive assertions using is(T matcher)

### DIFF
--- a/filter-processor/src/test/java/org/springframework/cloud/stream/module/filter/FilterProcessorIntegrationTests.java
+++ b/filter-processor/src/test/java/org/springframework/cloud/stream/module/filter/FilterProcessorIntegrationTests.java
@@ -18,13 +18,12 @@ package org.springframework.cloud.stream.module.filter;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertThat;
 import static org.springframework.cloud.stream.test.matcher.MessageQueueMatcher.receivesPayloadThat;
-
 import org.junit.Test;
 import org.junit.runner.RunWith;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.SpringApplicationConfiguration;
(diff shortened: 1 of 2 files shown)


filter-processor/src/test/java/org/springframework/cloud/stream/module/filter/FilterProcessorIntegrationTests.java
```diff
@@ -81,6 +81,56 @@ public void test() throws InterruptedException {
...
+			assertThat(collector.forChannel(channels.output()).poll(10, MILLISECONDS), is(nullValue(Message.class)));
+
+			channels.input().send(new GenericMessage<Object>("bazbar"));
+			channels.input().send(new GenericMessage<Object>("foobar"));
+			assertThat(collector.forChannel(channels.output()), receivesPayloadThat(not("bazbar")));
+			assertThat(collector.forChannel(channels.output()).poll(10, MILLISECONDS), is(nullValue(Message.class)));
+		}
+	}
+
+	@WebIntegrationTest("expression=payload=='baz'")
```
artembilan, line 116: I wonder why do we need these tests?
Looks like they check the SpEL stuff, not the `filter` processor...

There is really enough just one `expression` test on the matter.

transform-processor/src/test/java/org/springframework/cloud/stream/module/transform/TransformProcessorIntegrationTests.java
```diff
@@ -73,4 +74,60 @@ public void test() {
 			assertThat(collector.forChannel(channels.output()), receivesPayloadThat(is("HELLO")));
 		}
 	}
+
+	@WebIntegrationTest("expression=payload.toLowerCase()")
```
artembilan, line 78: The same here for all these new `expression` tests.

markpollack:
Hi.  Sorry Matt, I remember we had chatted before you started doing this, but Artem is correct - we already know that the range of functionality in a SpEL expressions works, so we don't need to have test coverage on SpEL expressions in the filter module.  Making sure that at least one expression is handled is ok,

PR #149, opened 2015-12-31
Introduce analytics API subproject

ericbottard:
- move FieldValueCounter related classes to a new submodule, so it can be depended upon by dataflow
- split FVCRepository in Reader+Writer, as does boot for its simple metrics
- add some read methods that are needed to displaying counters, rework the write API a bit to
  use a method that's more likely useful (full deletion rather than key deletion)

Fixes spring-cloud/spring-cloud-dataflow#247


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as f32e2a0

PR #154, opened 2016-01-07
initial commit of CloudFoundry Loggregator Spring Cloud Dataflow source

joshlong:
Provides [a Cloud Foundry Loggregator Source](http://docs.cloudfoundry.org/loggregator/architecture.html).

I'm not entirely sure how to test this. I have environment variables on my machine,  `CF_ORG`, `CF_SPACE`, `CF_USER`, `CF_PASSWORD`, and `CF_API` that work in this test. Also, I happen to know that I have an application called `configuration-service` available. 

I think the _right_ approach is to make sure to download `cf` in Travis, then use it to `cf push` a new simple Spring Boot CLI application called `foo-service`. We can have that app do almost nothing, just print things to the log. As soon as a single log entry's received, the test is a success. 

In my builds that use Travis and `cf`, the installation is as follows in the next script. `~/bin` and `~/.m2/repositories` could be cached through CI runs. 

```
#!/bin/bash


CF_USER=$1
CF_PASSWORD=$2
CF_ORG=$3
CF_SPACE=$4

function install_cf(){
    mkdir -p $HOME/bin
    curl -v -L -o cf.tgz 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.13.0&source=github-rel'
    tar zxpf cf.tgz
    mkdir -p $HOME/bin && mv cf $HOME/bin
}

function validate_cf(){

    cf  -v || install_cf

    export PATH=$PATH:$HOME/bin

    cf api https://api.run.pivotal.io
    cf auth $CF_USER $CF_PASSWORD
    cf target -o $CF_ORG -s $CF_SPACE
    cf apps
}

validate_cf

```

This source is useful because we can do smart things based on what our logs are telling us. I want to contribute an `ElasticSearchSource` and another that acts as a `syslog://` drain. Structured logs are a thing and for many applications they serve as an event plane and structured data. Tools like Kibana make analyzing this stuff as though it were schema dead simple.


trisberg:
I would think the best approach for the tests would be to not run integration tests that require access to a separate service/server by default. Similar to the gpfdist sink app here - https://github.com/jvalkeal/spring-cloud-stream-modules/commit/928b8206126a539bdeeb4302845f5a0497edce44#diff-6bcafa72fc357aeb2e6a8ac20294c7a5R68

trisberg:
Would be helpful to provide a small test app that could be deployed to CF, maybe a Boot Groovy app so it doesn't have to be compiled

joshlong:
Hi there @trisberg I've added a simple `it-test.sh` script and a `it-teardown.sh` script and updated the PR for your consideration

trisberg:
Merged via commit 020cd08d1bd6a14ee9bbbd2723064fe8b414d703

Some polishing/changes via commit 31ab8137433401de9a66a0937acd73d5c16b0004

PR #155, opened 2016-01-12
Add Jdbc Source

### LINKED ISSUE #86: Port JDBC as s-c-s source
As a developer, I'd like to move *jdbc* module from XD to s-c-s repo, so I can use it as source to build streaming pipeline. 
**Acceptance:**
* *jdbc* as source module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *jdbc* module with respective group/artifact/version detail

trisberg:
- Using same config options as XD version: --query, --split, --update, --maxRowsPerPoll, --fixedDelay, --maxMessages

Resolves #86

### DIFF
--- a/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
+++ b/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.stream.module.jdbc.source;
 
+import java.util.concurrent.TimeUnit;
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -34,8 +35,6 @@ import org.springframework.integration.dsl.support.Consumer;
 import org.springframework.integration.jdbc.JdbcPollingChannelAdapter;
 import org.springframework.integration.scheduling.PollerMetadata;
 import org.springframework.transaction.PlatformTransactionManager;
-
-import java.util.concurrent.TimeUnit;
 
 /**
  * A module that reads data from an RDBMS using JDBC and creates a payload with the data.
--- a/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
+++ b/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.jdbc.source;
+
+import javax.sql.DataSource;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.stream.annotation.Bindings;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Source;
+import org.springframework.cloud.stream.module.MaxMessagesProperties;
+import org.springframework.cloud.stream.module.PeriodicTriggerConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.integration.core.MessageSource;
+import org.springframework.integration.dsl.IntegrationFlow;
+import org.springframework.integration.dsl.IntegrationFlowBuilder;
+import org.springframework.integration.dsl.IntegrationFlows;
+import org.springframework.integration.dsl.SourcePollingChannelAdapterSpec;
+import org.springframework.integration.dsl.channel.MessageChannels;
+import org.springframework.integration.dsl.core.Pollers;
+import org.springframework.integration.dsl.support.Consumer;
+import org.springframework.integration.jdbc.JdbcPollingChannelAdapter;
+import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.transaction.PlatformTransactionManager;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A module that reads data from an RDBMS using JDBC and creates a payload with the data.
+ *
+ * @author Thomas Risberg
+ */
+@EnableBinding(Source.class)
+@EnableConfigurationProperties({JdbcSourceProperties.class, MaxMessagesProperties.class})
+@Import({PeriodicTriggerConfiguration.class})
+public class JdbcSourceConfiguration {
+
+	@Autowired
+	private JdbcSourceProperties properties;
+
+	@Autowired
+	DataSource dataSource;
+
+	@Autowired
+	PlatformTransactionManager transactionManager;
+
+	@Autowired
+	@Bindings(JdbcSourceConfiguration.class)
+	Source source;
+
+	@Bean
+	TaskExecutor taskExecutor() {
+		return new ThreadPoolTaskExecutor();
+	}
+
+	@Bean
+	@ConditionalOnProperty(value = "split", havingValue = "true", matchIfMissing = true)
+	public IntegrationFlow splitFlow() {
+		return IntegrationFlows.from("splitter")
+				.split("payload")
+				.channel(MessageChannels.executor(this.taskExecutor()))
+				.channel(source.output())
+				.get();
+	}
+
+	@Bean
+	public MessageSource<Object> jdbcMessageSource() {
+		JdbcPollingChannelAdapter jdbcPollingChannelAdapter =
+				new JdbcPollingChannelAdapter(dataSource, properties.getQuery());
+		jdbcPollingChannelAdapter.setMaxRowsPerPoll(properties.getMaxRowsPerPoll());
+		if (properties.getUpdate() != null) {
+			jdbcPollingChannelAdapter.setUpdateSql(properties.getUpdate());
+		}
+		return jdbcPollingChannelAdapter;
+	}
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.fixedRate(properties.getFixedDelay(), TimeUnit.SECONDS).
+				maxMessagesPerPoll(properties.getMaxMessages()).
+				transactional(transactionManager).get();
+	}
+
+	@Bean
+	public IntegrationFlow pollingFlow() {
+		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(jdbcMessageSource(),
+				new Consumer<SourcePollingChannelAdapterSpec>() {
+					@Override
+					public void accept(SourcePollingChannelAdapterSpec sourcePollingChannelAdapterSpec) {
+						sourcePollingChannelAdapterSpec
+								.autoStartup(false)
+								.poller(poller());
+					}
+				});
+		if (properties.isSplit()) {
+			flowBuilder.channel("splitter");
+		}
+		else {
+			flowBuilder.channel(source.output());
+		}
+		return flowBuilder.get();
+	}
+}
--- a/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
+++ b/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
@@ -19,24 +19,20 @@ package org.springframework.cloud.stream.module.jdbc.source;
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.stream.annotation.Bindings;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.messaging.Source;
 import org.springframework.context.annotation.Bean;
-import org.springframework.core.task.TaskExecutor;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.dsl.IntegrationFlow;
 import org.springframework.integration.dsl.IntegrationFlowBuilder;
 import org.springframework.integration.dsl.IntegrationFlows;
 import org.springframework.integration.dsl.SourcePollingChannelAdapterSpec;
-import org.springframework.integration.dsl.channel.MessageChannels;
 import org.springframework.integration.dsl.core.Pollers;
 import org.springframework.integration.dsl.support.Consumer;
 import org.springframework.integration.jdbc.JdbcPollingChannelAdapter;
 import org.springframework.integration.scheduling.PollerMetadata;
-import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.transaction.PlatformTransactionManager;
 
 import java.util.concurrent.TimeUnit;
@@ -62,21 +58,6 @@ public class JdbcSourceConfiguration {
 	@Autowired
 	@Bindings(JdbcSourceConfiguration.class)
 	private Source source;
-
-	@Bean
-	public TaskExecutor taskExecutor() {
-		return new ThreadPoolTaskExecutor();
-	}
-
-	@Bean
-	@ConditionalOnProperty(value = "split", havingValue = "true", matchIfMissing = true)
-	public IntegrationFlow splitFlow() {
-		return IntegrationFlows.from("splitter")
-				.split("payload")
-				.channel(MessageChannels.executor(this.taskExecutor()))
-				.channel(source.output())
-				.get();
-	}
 
 	@Bean
 	public MessageSource<Object> jdbcMessageSource() {
@@ -108,11 +89,9 @@ public class JdbcSourceConfiguration {
 					}
 				});
 		if (properties.isSplit()) {
-			flowBuilder.channel("splitter");
+			flowBuilder.split();
 		}
-		else {
-			flowBuilder.channel(source.output());
-		}
+		flowBuilder.channel(source.output());
 		return flowBuilder.get();
 	}
 }
--- a/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
+++ b/jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
@@ -24,10 +24,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cloud.stream.annotation.Bindings;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.messaging.Source;
-import org.springframework.cloud.stream.module.MaxMessagesProperties;
-import org.springframework.cloud.stream.module.PeriodicTriggerConfiguration;
 import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Import;
 import org.springframework.core.task.TaskExecutor;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.dsl.IntegrationFlow;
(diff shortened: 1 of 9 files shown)


jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+@EnableBinding(Source.class)
+@EnableConfigurationProperties({JdbcSourceProperties.class, MaxMessagesProperties.class})
+@Import({PeriodicTriggerConfiguration.class})
+public class JdbcSourceConfiguration {
+
+	@Autowired
+	private JdbcSourceProperties properties;
+
+	@Autowired
+	DataSource dataSource;
```
ericbottard: Any reason why this (and the ones below) are package protected?

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceProperties.java
```diff
@@ -0,0 +1,109 @@
...
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Holds configuration properties for the Jdbc Source module.
+ *
+ * @author Thomas Risberg
+ */
+@ConfigurationProperties
+public class JdbcSourceProperties {
```
ericbottard, line 29: Should add additional jsr303 validation where appropriate (and tests): maxRowsPerPoll, fixedDelay

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: No, just forgot to add private, will fix

jdbc-source/src/test/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceIntegrationTests.java
```diff
@@ -0,0 +1,171 @@
...
+	protected Source source;
+
+	@Autowired
+	protected JdbcOperations jdbcOperations;
+
+	@Autowired
+	protected MessageCollector messageCollector;
+
+	@IntegrationTest(value = {"query=select id, name from test order by id"})
+	public static class DefaultBehavior extends JdbcSourceIntegrationTests {
```
ericbottard: I believe these inner sub classes won't be run by surefire, because they're not named `XXXTests`

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
artembilan: Actually, it doesn't matter. The `@Autowired` doesn't care about modificator.
We need `private` only for the code quality.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A module that reads data from an RDBMS using JDBC and creates a payload with the data.
+ *
+ * @author Thomas Risberg
+ */
+@EnableBinding(Source.class)
+@EnableConfigurationProperties({JdbcSourceProperties.class, MaxMessagesProperties.class})
+@Import({PeriodicTriggerConfiguration.class})
```
artembilan: Please, pay attention to the #146 when we are fighting to reconsider and generalize any trigger stuff.
For example that `MaxMessagesProperties` which is out of your attention here, although you have faced an issue with the `-1` for the polling adapter.
So, I recommend to merge this one after that and particular reworking.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+	@Bean
+	TaskExecutor taskExecutor() {
+		return new ThreadPoolTaskExecutor();
+	}
+
+	@Bean
+	@ConditionalOnProperty(value = "split", havingValue = "true", matchIfMissing = true)
+	public IntegrationFlow splitFlow() {
+		return IntegrationFlows.from("splitter")
+				.split("payload")
```
artembilan: I guess your `payload` after the `JdbcPollingChannelAdapter` is `Collection` anyway:

```
List<?> payload = doPoll(this.sqlQueryParameterSource);
```

so, the default `.split()` (based on the `DefaultMessageSplitter`) meets your requirements.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+				.channel(MessageChannels.executor(this.taskExecutor()))
+				.channel(source.output())
```
artembilan: I don't think that we should worry about the `output` performance and parallelism.
The `source.output()` is pipe for the `Binder` - Redis, Kafka, Rabbit etc.

I mean that this `MessageChannels.executor()` is overhead.
And from here we even can get rid of this `IntegrationFlow` for the `split` option.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+		jdbcPollingChannelAdapter.setMaxRowsPerPoll(properties.getMaxRowsPerPoll());
+		if (properties.getUpdate() != null) {
+			jdbcPollingChannelAdapter.setUpdateSql(properties.getUpdate());
+		}
+		return jdbcPollingChannelAdapter;
+	}
+
+	@Bean
+	public PollerMetadata poller() {
+		return Pollers.fixedRate(properties.getFixedDelay(), TimeUnit.SECONDS).
```
artembilan: You say `fixedRate`, but use `getFixedDelay` options. 
Inconsistency.
Although we can forget this if we will be based on the #146 , where one `@MessageSource` will save us.
As well the `PollerMetadata` from there can be used for DSL definitions.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: OK, I'll get rid of it

jdbc-source/src/test/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceIntegrationTests.java
trisberg: You'r right, didn't think of that

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: I'll keep an eye on it

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: yup

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceProperties.java
artembilan, line 29: `query` is required...

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: Yes, I would agree. Inherited from XD where we do use fixed-delay poller, although I think it should have been fixed-rate in the first place. Will change to fixedDelay to stay consistent and then switch to #146 configs when available.

jdbc-source/src/test/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceIntegrationTests.java
```diff
@@ -0,0 +1,171 @@
...
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * Integration Tests for JdbcSource. Uses hsqldb as a (real) embedded DB.
+ *
+ * @author Thomas Risberg
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {JdbcSourceApplication.class, EmbeddedDataSourceConfiguration.class})
+@IntegrationTest({"server.port=-1"})
```
artembilan, line 51: Seems for me this doesn't work, because the particular implementation with the same `@IntegrationTest` overrides this value with that one.
That's why we use in other places `@WebIntegrationTest(randomPort = true)`.

No, this works, but with the default `8080` Tomcat port :smile:

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+						sourcePollingChannelAdapterSpec
+								.autoStartup(false)
+								.poller(poller());
+					}
+				});
+		if (properties.isSplit()) {
+			flowBuilder.split();
+		}
+		flowBuilder.channel(source.output());
+		return flowBuilder.get();
```
artembilan, line 87: This config looks cool :smile: 
Although with Java 8 we would be better...
Just rhetorical comment - no need to answer.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+	private Source source;
+
+	@Bean
+	public MessageSource<Object> jdbcMessageSource() {
+		JdbcPollingChannelAdapter jdbcPollingChannelAdapter =
+				new JdbcPollingChannelAdapter(dataSource, properties.getQuery());
+		jdbcPollingChannelAdapter.setMaxRowsPerPoll(properties.getMaxRowsPerPoll());
+		if (properties.getUpdate() != null) {
+			jdbcPollingChannelAdapter.setUpdateSql(properties.getUpdate());
+		}
```
artembilan: No reason to worry about the value. It is optional and `JdbcPollingChannelAdapter` takes care about its `null`.
I mean the `if` is redundant.
Of course, it is minor and we can live with this anyway :smile:

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	@Bean
+	public IntegrationFlow pollingFlow() {
+		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(jdbcMessageSource(),
+				new Consumer<SourcePollingChannelAdapterSpec>() {
+					@Override
+					public void accept(SourcePollingChannelAdapterSpec sourcePollingChannelAdapterSpec) {
+						sourcePollingChannelAdapterSpec
+								.autoStartup(false)
```
artembilan, line 79: Looks like something has been changed in the SCDF since XD, so we don't need to worry about `auto-startup` phase and the default behavior works as is.
I mean this `.autoStartup(false)` is redundant.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.integration.dsl.IntegrationFlowBuilder;
+import org.springframework.integration.dsl.IntegrationFlows;
+import org.springframework.integration.dsl.SourcePollingChannelAdapterSpec;
+import org.springframework.integration.dsl.core.Pollers;
+import org.springframework.integration.dsl.support.Consumer;
+import org.springframework.integration.jdbc.JdbcPollingChannelAdapter;
+import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.transaction.PlatformTransactionManager;
+
+import java.util.concurrent.TimeUnit;
```
artembilan: Something wrong with imports settings?
Or does SCS team use other style?..
imports

jdbc-source/src/test/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourcePropertiesTests.java
```diff
@@ -0,0 +1,150 @@
...
+	@Rule
+	public ExpectedException thrown= ExpectedException.none();
+
+	@Test
+	public void queryIsRequired() {
+		thrown.expect(BeanCreationException.class);
+		thrown.expectMessage("Field error in object 'target' on field 'query': rejected value [null]");
+		AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+		context.register(Conf.class);
+		context.refresh();
```
artembilan: `AnnotationConfigApplicationContext` is `Closeable`, so should be `close()`'d in the end, independently of test result.
At least STS barks here :smile:

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg: new IDE install, not configured properly, will fix

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg, line 87: I take cool ;)

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
trisberg, line 79: Yeah, not sure about this one, just followed the ftp-source which has it - https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/ftp-source/src/main/java/org/springframework/cloud/stream/module/ftp/source/FtpSource.java#L106

artembilan:
The #146 has been deprecated.
This stuff can be reworked to the `@PollableSource` or just `@Autowired PollerMetadata defaultPoller`, if that.

trisberg:
A bit confused how I would replace this:

```
    @Bean
    public PollerMetadata poller() {
        return Pollers.fixedDelay(properties.getFixedDelay(), TimeUnit.SECONDS).
                maxMessagesPerPoll(properties.getMaxMessages()).
                transactional(transactionManager).get();
    }
```

with

```
    @Autowired
    private PollerMetadata defaultPoller;
```

The fixedDelay default for the JDBC source is 5 and not 1 and how would I add in the transaction manager?

artembilan:
See `TriggerConfiguration`, there is a bean like:

```
@Bean
public PollerMetadata defaultPoller(Trigger trigger) {
```

You should use it like:

```
@Autowired 
@Qualifier("defaultPoller")
private PollerMetadata poller;
```

Since SCS already has its onw `PollerMetadata.DEFAULT_POLLER`, we can't just use `@Autowired` for that case.
Also see the `FileSource` exactly for the sample which you are looking for.

And yes: you should remove that your `poller()` altogether :smile:

trisberg:
had to add @Import(TriggerConfiguration.class) and works better now. Still confused about the transactional part - not needed?

artembilan:
Indeed. Let's think if we really need it here.
We read data from DB, may update it and send to the `output`. May do that for each record with `split`.
That's all: our work here is done. The data is send to the `Binder` over the network.
No anything else which may impact the process!
It is just done here on the `output` channel.

trisberg:
SO, are you saying that we don't need transactional? What would happen if one of the update statements failed?

artembilan:
> What would happen if one of the update statements failed?

We have there (`JdbcPollingChannelAdapter`) only one of them according this (JDBC Source module) our settings:

``` java
private Object poll() {
    List<?> payload = doPoll(this.sqlQueryParameterSource);
    if (payload.size() < 1) {
        payload = null;
    }
    if (payload != null && updateSql != null) {
        if (this.updatePerRow) {
            for (Object row : payload) {
                executeUpdateQuery(row);
            }
        }
        else {
            executeUpdateQuery(payload);
        }
    }
    return payload;
}
```

Where transaction, if that, is applied exactly around this `poll()`. Over the parent call stack, of course.
See `AbstractPollingEndpoint#createPoller()`  and `AbstractPollingEndpoint.Poller` class:

``` java
while (initialized && (maxMessagesPerPoll <= 0 || count < maxMessagesPerPoll)) {
    try {
            if (!pollingTask.call()) {
                break;
        }
    ....
}
```

That `pollingTask.call()` is transaction in our case. 
In other words the tx (or any poller's advice) is applied to the single `MessageSource` invocation only.

So, having that all I really don't see reason to have the TX for single SELECT/UPDATE. That works atomically on the RDBMS. You can reach better isolation with the `SELECT .... FOR UPDATE`, I think...

trisberg:
Rebased switched to use defaultPoller from TriggerConfiguration class

(this means that the default for fixedDelay is 1 instead of 5)

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceConfiguration.java
artembilan, line 79: I'll remove this on merge to avoid any similar discussions in the future.

jdbc-source/src/main/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceProperties.java
artembilan, line 29: Right, no reason to validate `maxRowsPerPoll`, because any `< 0` value is valid for the `JdbcPollingChannelAdapter` and we just pick up the full `SELECT` result on the output.

jdbc-source/src/test/java/org/springframework/cloud/stream/module/jdbc/source/JdbcSourceIntegrationTests.java
artembilan, line 51: I would like to see any feedback here.
I may not be right, but need to understand why you select this style of the config instead of `@WebIntegrationTest(randomPort = true)`.

Thank you!

ericbottard, line 51: IMO, this is to avoid creating a web server (which we don't need here). The tests are already taking quite some time, no need to add burden that is not needed

artembilan, line 51: Well, to avoid extra testing time we should avoid `@EnableAutoConfiguration`, which scans the classpath and populate a lot of stuff, which is, right, useless for the particular test-case.
But... Do we have other choice?

I'll check in the debug if `@IntegrationTest({"server.port=-1"})` helps to avoid starting Embedded Tomcat. 
Right now I don't believe: we have it in the CP, so `EmbeddedServletContainerAutoConfiguration` does its magic.

ericbottard, line 51: Well, we still want this to test the module as it would behave "in the real world"

artembilan, line 51: OK! Without `@WebIntegrationTest` we end up like:

```
DispatcherServletAutoConfiguration did not match
- @ConditionalOnClass classes found: org.springframework.web.servlet.DispatcherServlet (OnClassCondition)
- not a web application (OnWebApplicationCondition)
```

But the same result we get even without this noisy top-level `@IntegrationTest`.
In this scenario it does not have any effect: independently of that `server.port` value we don't step into the `ServerProperties` at all.

What is funny here that with the `@WebIntegrationTest({"server.port=-1"})` we end up with the:

```
2016-01-19 13:15:38.946  INFO 8796 --- [           main] s.b.c.e.t.TomcatEmbeddedServletContainer : Tomcat started on port(s): -1 (http)
```

Although the Boot Docs say:

> To switch off the HTTP endpoints completely, but still create a WebApplicationContext, use server.port=-1 (this is sometimes useful for testing).

However we have the protection code in the `TomcatEmbeddedServletContainerFactory`:

```
protected void customizeConnector(Connector connector) {
        int port = (getPort() >= 0 ? getPort() : 0);
        connector.setPort(port);
```

Maybe that "endpoints disabling" doesn't relate to the Embedded Servlet Container...
Although yes: it isn't started:

```
return new TomcatEmbeddedServletContainer(tomcat, getPort() >= 0);
....
public void start() throws EmbeddedServletContainerException {
    try {
        addPreviouslyRemovedConnectors();
        Connector connector = this.tomcat.getConnector();
        if (connector != null && this.autoStart) {
            startConnector(connector);
              }
```

So, I guess we can remove this annotation from here, but for other tests when we would like to share some options for all tests, like we do in the `CassandraSinkIntegrationTests`, we don't have choice unless use `@WebIntegrationTest`. Otherwise the downstream `@IntegrationTest` overrides any properties from the upstream.

But I agree that it would be good to find some compromise between real world testing and the entire Spring Boot noise in the particular test.

artembilan:
In addition the `sources.adoc` article must be there.
I'll show you polishing on the matter.

artembilan:
Here you are the polishing for your confirmation: https://github.com/artembilan/spring-cloud-stream-modules/commit/1fbb15bc17f2b1e4697ceecd0ab92ce7af49c28d

artembilan:
@trisberg , does your "LGTM" mean that I can merge ?

trisberg:
Yes, go for it :)

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/48f8c2df4c9a8a05cab60c7170622ded98997823

artembilan:
Also see https://github.com/spring-cloud/spring-cloud-dataflow/commit/e2c10359e769c7a03626ca33449ecd2e04b1e122

PR #156, opened 2016-01-13
Add module used in integration tests

ericbottard:
This adds a module that can be told to stall or fail deployment

### DIFF
--- a/integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
+++ b/integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
@@ -19,8 +19,9 @@ package org.springframework.cloud.stream.module.it;
 import javax.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.cloud.stream.messaging.Processor;
 
 /**
@@ -29,17 +30,18 @@ import org.springframework.cloud.stream.messaging.Processor;
  * @author Eric Bottard
  */
 @EnableBinding(Processor.class)
+@EnableConfigurationProperties(IntegrationTestProcessorProperties.class)
 public class IntegrationTestProcessor {
 
 	@Autowired
 	private IntegrationTestProcessorProperties properties;
 
-	@Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
-	private int instanceIndex;
+	@Autowired
+	private ChannelBindingServiceProperties channelBindingServiceProperties;
 
 	@PostConstruct
 	public void init() throws InterruptedException {
-		if (properties.getMatchInstances().isEmpty() || properties.getMatchInstances().contains(instanceIndex)) {
+		if (properties.getMatchInstances().isEmpty() || properties.getMatchInstances().contains(instanceIndex())) {
 			Thread.sleep(properties.getInitDelay());
 			if (properties.getKillDelay() >= 0) {
 				new Thread() {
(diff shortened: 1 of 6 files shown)


mbogoevici:
I'm wondering if this wouldn't better be added as an artifact to `spring-cloud-dataflow`, besides the TCK. Thus it could be versioned/referenced/released together with the TCK itself. Also, it doesn't seem to be something that we'd want to discover automatically (if/when we decide to support that in admin).

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Eric Bottard
+ */
+@EnableBinding(Processor.class)
+public class IntegrationTestProcessor {
+
+	@Autowired
+	private IntegrationTestProcessorProperties properties;
+
+	@Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
+	private int instanceIndex;
```
mbogoevici: Would there be any value in injecting ChannelBindingServiceProperties and getting the instance index from there? Just for the sake of avoiding duplicating the logic there (in case it changes for whatever reason)?

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessorApplication.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.it;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+/**
+ * Created by ericbottard on 12/01/16.
```
ilayaperumalg, line 23: Can fix the author tag while merging.

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessorProperties.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Configuration properties for the IntegrationTestProcessor module.
+ *
+ * @author Eric Bottard
+ */
+public class IntegrationTestProcessorProperties {
```
ilayaperumalg, line 30: This needs to be annotated as `@ConfigurationProperties`

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Processor;
+
+/**
+ * A module that can misbehave, useful for integration testing of module deployers.
+ *
+ * @author Eric Bottard
+ */
+@EnableBinding(Processor.class)
+public class IntegrationTestProcessor {
```
ilayaperumalg, line 34: I believe we need to `@EnableConfigurationProperties` for `IntegrationTestProcessorProperties` so that it can be autowired.

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
ilayaperumalg: Yeah, we can just auto-wire `ChannelBindingServiceProperties` and refer `instanceIndex` from it.

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessorProperties.java
ericbottard, line 30: Interesting. It does work without it.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and rebased as f0a3cb8.

PR #157, opened 2016-01-18
GH-89: RabbitMQ Source

### LINKED ISSUE #89: Port Rabbit as s-c-s source
As a s-c-d developer, I'd like to move *rabbit* module from XD to s-c-s repo, so I can use it as *source* to build streaming pipeline.

Consider addressing [XD-2790](https://jira.spring.io/browse/XD-2790)

**Acceptance:**
* *rabbit* as source module is available in *spring-cloud-stream-modules* repo
* maven repo includes *rabbit* module with respective group/artifact/version detail

garyrussell:
Resolves #89

### DIFF
--- a/rabbit-source/pom.xml
+++ b/rabbit-source/pom.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>rabbit-source</artifactId>
+	<packaging>jar</packaging>
+	<name>rabbit-source</name>
+	<description>Rabbit Source stream module</description>
+
+	<parent>
+		<groupId>org.springframework.cloud.stream.module</groupId>
+		<artifactId>spring-cloud-stream-modules</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<start-class>org.springframework.cloud.stream.module.rabbit.source.RabbitSourceApplication</start-class>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-amqp</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
+
+</project>
--- a/rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
+++ b/rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.rabbit.source;
+
+import org.aopalliance.aop.Advice;
+
+import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.rabbit.config.RetryInterceptorBuilder;
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
+import org.springframework.amqp.rabbit.retry.RejectAndDontRequeueRecoverer;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.amqp.RabbitProperties;
+import org.springframework.boot.autoconfigure.amqp.RabbitProperties.Listener;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.stream.annotation.Bindings;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Source;
+import org.springframework.context.annotation.Bean;
+import org.springframework.integration.amqp.inbound.AmqpInboundChannelAdapter;
+import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
+import org.springframework.retry.interceptor.RetryOperationsInterceptor;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * A source module that receives data from RabbitMQ.
+ *
+ * @author Gary Russell
+ */
+@EnableBinding(Source.class)
+@EnableConfigurationProperties(RabbitSourceProperties.class)
+public class RabbitSource {
+
+	@Autowired
+	@Bindings(RabbitSource.class)
+	private Source channels;
+
+	@Autowired
+	private RabbitProperties rabbitProperties;
+
+	@Autowired
+	private RabbitSourceProperties properties;
+
+	@Autowired(required = false)
+	private ConnectionFactory rabbitConnectionFactory;
+
+	@Bean
+	public SimpleMessageListenerContainer container() {
+		SimpleMessageListenerContainer container = new SimpleMessageListenerContainer(this.rabbitConnectionFactory);
+		container.setAutoStartup(false);
+		Listener listenerProperties = this.rabbitProperties.getListener();
+		AcknowledgeMode acknowledgeMode = listenerProperties.getAcknowledgeMode();
+		if (acknowledgeMode != null) {
+			container.setAcknowledgeMode(acknowledgeMode);
+		}
+		Integer concurrency = listenerProperties.getConcurrency();
+		if (concurrency != null) {
+			container.setConcurrentConsumers(concurrency);
+		}
+		Integer maxConcurrency = listenerProperties.getMaxConcurrency();
+		if (maxConcurrency != null) {
+			container.setMaxConcurrentConsumers(maxConcurrency);
+		}
+		Integer prefetch = listenerProperties.getPrefetch();
+		if (prefetch != null) {
+			container.setPrefetchCount(prefetch);
+		}
+		Integer transactionSize = listenerProperties.getTransactionSize();
+		if (transactionSize != null) {
+			container.setTxSize(transactionSize);
+		}
+		Boolean defaultRequeueRejected = this.properties.getRequeue();
+		if (defaultRequeueRejected != null) {
+			container.setDefaultRequeueRejected(defaultRequeueRejected);
+		}
+		Boolean channelTransacted = this.properties.getChannelTransacted();
+		if (channelTransacted != null) {
+			container.setChannelTransacted(channelTransacted);
+		}
+		String queues = this.properties.getQueues();
+		Assert.state(StringUtils.hasText(queues), "At least one queue is required");
+		container.setQueueNames(StringUtils.commaDelimitedListToStringArray(queues));
+		if (this.properties.isEnableRetry()) {
+			container.setAdviceChain(new Advice[] { rabbitSourceRetryInterceptor() });
+		}
+		return container;
+	}
+
+	@Bean
+	public AmqpInboundChannelAdapter adapter() {
+		AmqpInboundChannelAdapter adapter = new AmqpInboundChannelAdapter(container());
+		adapter.setAutoStartup(false);
+		adapter.setOutputChannel(channels.output());
+		DefaultAmqpHeaderMapper headerMapper = new DefaultAmqpHeaderMapper();
+		headerMapper.setRequestHeaderNames(
+				StringUtils.commaDelimitedListToStringArray(this.properties.getMappedRequestHeaders()));
+		adapter.setHeaderMapper(headerMapper);
+		return adapter;
+	}
+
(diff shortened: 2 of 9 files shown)


rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
```diff
@@ -0,0 +1,126 @@
...
+	private Source channels;
+
+	@Autowired
+	private RabbitProperties rabbitProperties;
+
+	@Autowired
+	private RabbitSourceProperties properties;
+
+	@Autowired(required = false)
+	private ConnectionFactory rabbitConnectionFactory;
```
artembilan, line 60: `required = false`?
How can we live without `connectionFactory`?

```
public void afterPropertiesSet() {
    Assert.notNull(this.connectionFactory, "ConnectionFactory is required");
}
```

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
```diff
@@ -0,0 +1,126 @@
...
+		if (this.properties.isEnableRetry()) {
+			container.setAdviceChain(new Advice[] { rabbitSourceRetryInterceptor() });
+		}
+		return container;
+	}
+
+	@Bean
+	public AmqpInboundChannelAdapter adapter() {
+		AmqpInboundChannelAdapter adapter = new AmqpInboundChannelAdapter(container());
+		adapter.setAutoStartup(false);
```
artembilan: Is this still necessary?
As well as the same for the `container()` above:

```
this.messageListenerContainer = listenerContainer;
this.messageListenerContainer.setAutoStartup(false);
```

M-m-m. Minor, of course, if there is someone who cares about all these our `Lifecycle`s...

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
```diff
@@ -0,0 +1,150 @@
...
+
+	/**
+	 * Whether the channel is transacted.
+	 */
+	private Boolean channelTransacted;
+
+	/**
+	 * The queues to which the source will listen for messages.
+	 */
+	private String queues;
```
artembilan: Haven't you tried with `String[]`?
I agree about `"At least one queue is required"`, but seems for me the `ConversionService` can do its magic for us here.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
```diff
@@ -0,0 +1,150 @@
...
+
+	/**
+	 * Retry backoff multiplier when retry is enabled.
+	 */
+	private double retryMultiplier = 2.0;
+
+	/**
+	 * The maximum delivery attempts when retry is enabled.
+	 */
+	private int maxAttempts = 3;
```
artembilan, line 70: I wonder if we should worry about negative values for this options.
I mean something like `@Min(0)` on their getters.

spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
```diff
@@ -102,6 +102,48 @@ $$messageSize$$:: $$the size of message to send$$ *($$Integer$$, `1000`)*
...
+
+The **$$rabbit$$** $$source$$ has the following options:
+
+$$enableRetry$$:: $$enable retry; when retries are exhausted the message will be rejected; message disposition will depend on dead letter configuration$$ *($$boolean$$, default: `false`)*
+$$initialRetryInterval$$:: $$initial interval between retries$$ *($$int$$, default: `1000`)*
+$$mappedRequestHeaders$$:: $$request message header names to be propagated to/from the adpater/gateway$$ *($$String$$, default: `STANDARD_REQUEST_HEADERS`)*
+$$maxAttempts$$:: $$maximum delivery attempts$$ *($$int$$, default: `3`)*
+$$maxConcurrency$$:: $$the maximum number of consumers$$ *($$int$$, default: `1`)*
+$$maxRetryInterval$$:: $$maximum retry interval$$ *($$int$$, default: `30000`)*
+$$queues$$:: $$the queue(s) from which messages will be received$$ *($$String$$, default: `<stream name>`)*
```
artembilan: There is no default in the `RabbitSourceProperties`

spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
```diff
@@ -102,6 +102,48 @@ $$messageSize$$:: $$the size of message to send$$ *($$Integer$$, `1000`)*
...
+$$requeue$$:: $$whether rejected messages will be requeued by default$$ *($$boolean$$, default: `true`)*
```
artembilan, line 124: M-m-m. Yes it is like `defaultRequeueRejected = true;` in the `SimpleMessageListenerContainer`, but `RabbitSourceProperties` doesn't reflect that.
Seems for me the CLI can show all properties and its defaults, but not from the target objects.

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceTests.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Tests for RabbitSource.
+ *
+ * @author Gary Russell
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = { RabbitSourceApplication.class, RabbitSourceTests.Config.class })
+@DirtiesContext
+@WebIntegrationTest(randomPort = true, value = "port = 0")
```
artembilan: What is the `port = 0`? Who is responsible for this option? 
I don't see such a property in the `RabbitSourceProperties`

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceTests.java
```diff
@@ -0,0 +1,153 @@
...
+public abstract class RabbitSourceTests {
+
+	@Rule
+	public RabbitTestSupport rabbitAvailable;
```
artembilan, line 62: I'm curious: why this is in the `internal`?

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
garyrussell, line 60: This is left over from when I was trying to use separate config from the boot config. Needs to be removed.

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceTests.java
garyrussell, line 62: See the pom description for internal.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
garyrussell: No, I forgot - @mbogoevici fixed it with the phases the bindings are started.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
garyrussell, line 70: Not really necessary, I think; anything < 1 is treated as 0.

spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
garyrussell, line 124: Yes - it's a dilemma - Seems wrong to replicate the default in the properties but I suppose it won't hurt.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
artembilan, line 70: Yeah... I won't mind. But especially this one will make undesired, "unretryable" affect:

```
public boolean canRetry(RetryContext context) {
    Throwable t = context.getLastThrowable();
    return (t == null || retryForException(t)) && context.getRetryCount() < maxAttempts;
}
```

:smile:

spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
```diff
@@ -102,6 +102,48 @@ $$messageSize$$:: $$the size of message to send$$ *($$Integer$$, `1000`)*
...
+$$retryMultiplier$$:: $$retry interval multiplier$$ *($$double$$, default: `2.0`)*
+$$transacted$$:: $$true if the channel is to be transacted$$ *($$boolean$$, default: `false`)*
+
+Also see the http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#common-application-properties[Spring Boot Documentation]
```
artembilan: Yeah... I wouldn't use `htmlsingle` because it loads the whole docs. And it doesn't that enough slowly.
The http://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html would be better.

garyrussell:
Pushed

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
garyrussell, line 70: It not only won't retry, it won't try at all :smile:

garyrussell:
Repushed with fixed boot link.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
```diff
@@ -0,0 +1,150 @@
...
+	public boolean getTransacted() {
+		return transacted;
+	}
+
+	public void setTransacted(boolean transacted) {
+		this.transacted = transacted;
+	}
+
+	@NotNull(message = "At least one queue is required")
+	public String[] getQueues() {
```
artembilan, line 95: I think `@Size(min = 1)` would be better.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
```diff
@@ -0,0 +1,119 @@
...
+		}
+		Integer transactionSize = listenerProperties.getTransactionSize();
+		if (transactionSize != null) {
+			container.setTxSize(transactionSize);
+		}
+		container.setDefaultRequeueRejected(this.properties.getRequeue());
+		container.setChannelTransacted(this.properties.getTransacted());
+		String[] queues = this.properties.getQueues();
+		Assert.state(queues.length > 0, "At least one queue is required");
+		Assert.noNullElements(queues, "queues cannot have null elements");
```
artembilan, line 90: Shouldn't we utilize any `Assert` as a validation functions in the `@ConfigurationProperties`?

From other side I'm sure there are so much `Assert`s downstream which allows us do not care here on the SCS level with those extra validations...

Or am I missing anything and we always should validate any suspicious options...

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceTests.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Tests for RabbitSource.
+ *
+ * @author Gary Russell
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = { RabbitSourceApplication.class, RabbitSourceTests.Config.class })
+@DirtiesContext
+@WebIntegrationTest(randomPort = true)
```
artembilan: Please, take a look to our discussion on the matter: https://github.com/spring-cloud/spring-cloud-stream-modules/pull/155#discussion_r49519333

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSource.java
garyrussell, line 90: These tests should really be in the SMLC - they are there for `addQueueNames` but not `setQueueNames`. I'll open a JIRA over there; we can remove these later.

rabbit-source/src/main/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceProperties.java
garyrussell, line 95: Needs `@NotNull` as well.

garyrussell:
Pushed

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceTests.java
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+	}
+
+//	@IntegrationTest({ "enableRetry = false" })
+//	public static class NoQueuesTests extends RabbitSourceTests {
+//
+//		@Test
+//		public void test() throws Exception {
+//		}
```
artembilan: How about this instead:

``` java
@Test
public void testNoQueues() throws Exception {
    try {
        SpringApplication.run(RabbitSourceApplication.class, "enableRetry = false");
        fail("BeanCreationException expected");
    }
    catch (Exception e) {
        assertThat(e, instanceOf(BeanCreationException.class));
        assertThat(e.getMessage(), containsString("queue(s) are required"));
    }
}
```

?

artembilan: Of course, the `setWebEnvironment(false)` should be in use :smile:

garyrussell:
Pushed

rabbit-source/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+	<properties>
+		<start-class>org.springframework.cloud.stream.module.rabbit.source.RabbitSourceApplication</start-class>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+		</dependency>
```
artembilan, line 24: `<scope>test</scope>`?
From other side the parent pom (`spring-cloud-stream-modules`) says that this dependency is in use.
So, it is redundant here in the particular module.

artembilan, line 24: That's all.
Fixing on merge...

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/baeb8606ecaff5232d80d9507d6d019ba2346049

Issue #164, opened 2016-01-22
Rabbit source module tests need @Rule for when rabbit is not available


ericbottard:
Currently, one can't build the modules project if rabbit-server is not running locally.

Those tests should be made conditional on the availability of rabbit

artembilan:
@ericbottard , looks like it is here already:
```
/**
 * Abstract base class for JUnit {@link Rule}s that detect the presence of some external resource. If the resource is
 * indeed present, it will be available during the test lifecycle through {@link #getResource()}. If it is not, tests
 * will either fail or be skipped, depending on the value of system property {@value #SCS_EXTERNAL_SERVERS_REQUIRED}.
 *
 * @author Eric Bottard
 * @author Gary Russell
 */
public abstract class AbstractExternalResourceTestSupport<R> implements TestRule {
```
See its `failOrSkip()` for the proper usage.

And those Rabbit Source tests use `RabbitTestSupport` `@Rule`.

Or do you have anything else in mind?

ericbottard:
Weird, I assumed it was not there, as when I ran the build with rabbit off it failed, and it worked as soon as I turned rabbit on.

I'll look into it

ericbottard:
No can do, it fails consistently for me on `PropertiesPopulatedTests` when rabbit is not running

artembilan:
Oh! 
I think the issue that we start Spring Boot with all the stuff before the reaching `TestRule`.
Try with:
```
@ClassRule
public static RabbitTestSupport rabbitAvailable = new RabbitTestSupport();
```

ericbottard:
As soon as you properly initialize the rule, it behaves better :)

ericbottard:
no need for Class rule, we just need to initialize it...

artembilan:
:smile: 

But with the `@ClassRule` we avoid some Boot noise in logs:
```
org.springframework.amqp.AmqpConnectException: java.net.ConnectException: Connection refused: connect
	at org.springframework.amqp.rabbit.support.RabbitExceptionTranslator.convertRabbitAccessException(RabbitExceptionTranslator.java:58) ~[spring-rabbit-1.5.2.RELEASE.jar:na]
```
and tests run faster.

PR #165, opened 2016-01-22
Properly initialize the rabbit @Rule

ericbottard:
Fixes spring-cloud/spring-cloud-stream-modules#164

### DIFF
--- a/rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceInvalidConfigTests.java
+++ b/rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceInvalidConfigTests.java
@@ -21,7 +21,7 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.junit.Assert.fail;
 
-import org.junit.Rule;
+import org.junit.ClassRule;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


artembilan:
The `RabbitSourceInvalidConfigTests` should be fixed as well.

ericbottard:
> The RabbitSourceInvalidConfigTests should be fixed as well.

Actually, none of those get a chance to hit rabbit

ericbottard:
Updated my PR

rabbit-source/src/test/java/org/springframework/cloud/stream/module/rabbit/source/RabbitSourceInvalidConfigTests.java
```diff
@@ -38,8 +38,8 @@
  */
 public class RabbitSourceInvalidConfigTests {
 
-	@Rule
-	public RabbitTestSupport rabbitAvailable;
+	@ClassRule
+	public static RabbitTestSupport rabbitAvailable = new RabbitTestSupport();
```
artembilan, line 42: Well, looking closely to this test, I'd say that we really don't need this rule here.
Our `Config` below is based only on the `RabbitSourceProperties`.

@ericbottard , you are right: it doesn't have hooks to try to access to Rabbit.

I can address that on merge.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/9a52ea666f48b843e4f92eea76037eafc0b50c84

PR #169, opened 2016-01-28
Some fixes and polishing

artembilan:
- Upgrade to SI Java DSL 1.1.2
- Introduce `SourcePayloadProperties` to move `payload` option outside of any `Source` module.
  Right now it makes sense only for the `TriggerSource`.
- Make some tests ignorable on Windows, since they are not intended to pass there.
- Add `server.port=-1` for the `CassandraSinkIntegrationTests` to make it a bit faster bypassing Tomcat start.
- Close context in the `CassandraSinkPropertiesTests`.
- Remove redundant usage of `MaxMessagesProperties` in the Source modules since it is picked up by the `TriggerConfiguration`.

### DIFF
--- a/common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
+++ b/common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
@@ -18,17 +18,24 @@ package org.springframework.cloud.stream.module.ftp;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.fail;
 
+import org.junit.Assume;
 import org.junit.Test;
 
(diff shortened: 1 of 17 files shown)


common/spring-cloud-stream-modules-common-configuration/src/test/java/org/springframework/cloud/stream/module/ftp/FtpSessionFactoryPropertiesTests.java
```diff
@@ -82,6 +89,23 @@ public void passwordCanBeCustomized() {
...
+		Assume.assumeFalse(System.getProperty("os.name").startsWith("Windows"));
+		try (AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext()) {
+			context.register(Conf.class);
+			context.refresh();
+			fail("exception expected");
+		}
+		catch (Exception e) {
+			assertThat(e.getCause(), instanceOf(BindException.class));
+			BindException bindException = (BindException) e.getCause();
+			FieldError fieldError = (FieldError) bindException.getAllErrors().get(0);
```
ericbottard, line 103: This may fail later if we add other constraints

garyrussell:
LGTM; polishing here https://github.com/garyrussell/spring-cloud-stream-modules/commit/42863c86224b4a413d6b9335aca77145f8a63dfb to address @ericbottard 's concern.

garyrussell:
Merged as 7f094f306e0871fe6e1bb4b97e636c3d7f96bc16

PR #170, opened 2016-01-29
Kafka source #90

### LINKED ISSUE #90: Port Kafka as s-c-s source
As a s-c-d developer, I'd like to move *kafka* module from XD to s-c-s repo, so I can use it as *source* to build streaming pipeline.
**Acceptance:**
* *kafka* as source module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *kafka* module with respective group/artifact/version detail

sobychacko:
resolves #90


kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceApplication.java
```diff
@@ -0,0 +1,34 @@
+/*
```
markfisher, line 1: some funky formatting for this header comment

sobychacko, line 1: Just updated the PR with formatting changes.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Properties related to connecting to Kafka broker.
+ *
+ * @author Soby Chacko
+ */
+@ConfigurationProperties
+public class KafkaConfigurationProperties {
```
ilayaperumalg, line 36: There are few other configuration properties such as `concurrency`, `maxFetch`, `queueSize` and `stopTimeout` of `KafkaMessageListenerContainer` can be defined here.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,153 @@
...
+		String keyDecoder = properties.getKeyDecoder();
+		if (keyDecoder == null) {
+			return new DefaultDecoder(new VerifiableProperties());
+		}
+		Class keyDecoderClass = Class.forName(keyDecoder);
+		return (Decoder) keyDecoderClass.newInstance();
+	}
+
+	private Decoder<?> getPayloadDecoder() throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+		String payloadDecoder = properties.getKeyDecoder();
```
ilayaperumalg: properties.getPayloadDecoder() here.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,153 @@
...
+	public IntegrationFlow fromKafka() throws IllegalAccessException, InstantiationException, ClassNotFoundException {
+
+		return IntegrationFlows.from(Kafka.messageDriverChannelAdapter(container())
+				.keyDecoder(getKeyDecoder())
+				.payloadDecoder(getPayloadDecoder()))
+				.channel(source.output())
+				.get();
+	}
+
+	private Decoder<?> getKeyDecoder() throws ClassNotFoundException, IllegalAccessException, InstantiationException {
```
ilayaperumalg: Can we refactor both of these `getKeyDecoder()` and `getPayloadDecoder()` to `getDecoder(String decoderType)` as both look functionally the same?

pom.xml
```diff
@@ -110,6 +111,11 @@
 				</exclusions>
 			</dependency>
 			<dependency>
+				<groupId>org.springframework.integration</groupId>
```
ilayaperumalg: I am not sure why do we need this in parent POM. May be this can be removed with just the version tag on the kafka source POM.

kafka-source/README.adoc
```diff
@@ -0,0 +1,24 @@
...
+== Build
+
+```
+$> mvn package
+```
+
+== Run
+
+```
+$> java -jar target/kafka-source-${version}-exec.jar
```
ilayaperumalg, line 23: It would be good to include the doc on how would someone specify the `partitions` and `initialOffsets` options when running the module here.

For instance with options like these:

```
--topics=test123,test234 --partitions.test123=1,2 --partitions.test234=1
```

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,153 @@
...
+		return container;
+	}
+
+	@Bean
+	public IntegrationFlow fromKafka() throws IllegalAccessException, InstantiationException, ClassNotFoundException {
+
+		return IntegrationFlows.from(Kafka.messageDriverChannelAdapter(container())
+				.keyDecoder(getKeyDecoder())
+				.payloadDecoder(getPayloadDecoder()))
+				.channel(source.output())
```
ilayaperumalg: Can this be simplified to use the channel name `Source.OUTPUT` directly instead of `source.output()` since the `IntegrationFlows` builder supports specifying the channel name in this method?

ericbottard: IMO, using the channel reference is better. Even if using `Source.OUTPUT` would be refactoring-friendly, it's just some attribute of the channel. What we want to convey is that we want _that_ channel, wouldn't you agree?

ilayaperumalg: I just tried running the source module exec with the `topics` option set and I see the following when using `source.output()` here. But this works ok when having the channel name `Source.OUTPUT`. Looks like it has got something to do with the channel name being assigned for the module output channel `output` vs 'application.fromKafka.channel#0'.?

```
org.springframework.messaging.MessageDeliveryException: Dispatcher has no subscribers for channel 'application.fromKafka.channel#0'.; nested exception is org.springframework.integration.MessageDispatchingException: Dispatcher has no subscribers
    at org.springframework.integration.channel.AbstractSubscribableChannel.doSend(AbstractSubscribableChannel.java:81) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:442) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:392) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:115) ~[spring-messaging-4.2.4.RELEASE.jar!/:4.2.4.RELEASE]
    at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:45) ~[spring-messaging-4.2.4.RELEASE.jar!/:4.2.4.RELEASE]
    at org.springframework.messaging.core.AbstractMessageSendingTemplate.send(AbstractMessageSendingTemplate.java:105) ~[spring-messaging-4.2.4.RELEASE.jar!/:4.2.4.RELEASE]
    at org.springframework.integration.endpoint.MessageProducerSupport.sendMessage(MessageProducerSupport.java:105) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter.access$300(KafkaMessageDrivenChannelAdapter.java:43) ~[spring-integration-kafka-1.3.0.RELEASE.jar!/:na]
    at org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter$AutoAcknowledgingChannelForwardingMessageListener.doOnMessage(KafkaMessageDrivenChannelAdapter.java:171) ~[spring-integration-kafka-1.3.0.RELEASE.jar!/:na]
    at org.springframework.integration.kafka.listener.AbstractDecodingMessageListener.onMessage(AbstractDecodingMessageListener.java:50) ~[spring-integration-kafka-1.3.0.RELEASE.jar!/:na]
    at org.springframework.integration.kafka.listener.QueueingMessageListenerInvoker$KafkaMessageDispatchingSubscriber.onNext(QueueingMessageListenerInvoker.java:221) [spring-integration-kafka-1.3.0.RELEASE.jar!/:na]
    at org.springframework.integration.kafka.listener.QueueingMessageListenerInvoker$KafkaMessageDispatchingSubscriber.onNext(QueueingMessageListenerInvoker.java:209) [spring-integration-kafka-1.3.0.RELEASE.jar!/:na]
    at reactor.core.processor.util.RingBufferSubscriberUtils.route(RingBufferSubscriberUtils.java:67) [reactor-core-2.0.7.RELEASE.jar!/:na]
    at reactor.core.processor.RingBufferProcessor$BatchSignalProcessor.run(RingBufferProcessor.java:789) [reactor-core-2.0.7.RELEASE.jar!/:na]
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_20]
    at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_20]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_20]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [na:1.8.0_20]
    at java.lang.Thread.run(Thread.java:745) [na:1.8.0_20]
Caused by: org.springframework.integration.MessageDispatchingException: Dispatcher has no subscribers
    at org.springframework.integration.dispatcher.UnicastingDispatcher.doDispatch(UnicastingDispatcher.java:153) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.integration.dispatcher.UnicastingDispatcher.dispatch(UnicastingDispatcher.java:120) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    at org.springframework.integration.channel.AbstractSubscribableChannel.doSend(AbstractSubscribableChannel.java:77) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
    ... 18 common frames omitted
```

mbogoevici: +1 on using channel reference

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
sobychacko, line 36: @ilayaperumalg I am adding these 4 properties. Any other properties that you can think of? I think we can start with these and add more later on as the need arises.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
sobychacko: @ilayaperumalg the decoders may not be the same. It was a typo that we used the KeyDecoder for both key and payload. Thanks a lot for catching this. We have separate properties for key and payload decoders in the KafkaConfigurationProperties class. I will change the code to reflect that. 
For example, Key could be a LongDecoder and the payload could be based on String. With that said, we can refactor some common elements between these two methods.

sobychacko: @ilayaperumalg with that said, I think we can re-factor some common elements from these two methods.

pom.xml
sobychacko: @ilayaperumalg sure thing - i pulled it from the parent and put it in the module itself.

kafka-source/README.adoc
sobychacko, line 23: Good point, will address this.

sobychacko:
Addressed the feedback of the review. I haven't changed the output channel from reference to Source.OUTPUT as we probably need to discuss this a bit further it seems like.

sobychacko:
@mbogoevici Few more changes need to be committed for the source.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
ilayaperumalg, line 36: @sobychacko yeah, and those are the only properties I found missing as well.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,158 @@
...
+	private Source source;
+
+	@Autowired
+	private KafkaConfigurationProperties properties;
+
+	@Bean
+	public KafkaMessageListenerContainer container() {
+
+		Assert.isTrue(!ObjectUtils.isEmpty(properties.getTopics()) || !ObjectUtils.isEmpty(properties.getPartitions()),
+				"Either a list of topics OR a topic=<comma separated partitions> must be provided");
```
artembilan: The `Assert` must like `validation` in the `KafkaConfigurationProperties`.
From other side this message doesn't reflect `topics/partitions` reality.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,158 @@
...
+		ZookeeperConnect zookeeperConnect = new ZookeeperConnect();
+		zookeeperConnect.setZkConnect(properties.getZkConnect());
+		zookeeperConnect.setZkConnectionTimeout(properties.getZkConnectionTimeout());
+		zookeeperConnect.setZkSessionTimeout(properties.getZkSessionTimeout());
+		zookeeperConnect.setZkSyncTime(properties.getZkSyncTime());
+		return zookeeperConnect;
+	}
+
+	@Bean
+	public OffsetManager kafkaNativeOffsetManager(ConnectionFactory kafkaConnectionFactory) {
```
artembilan: I don't see the `ConnectionFactory` `@Bean`.
Who populates it?
What have I missed? Why then `container()` bean doesn't use the same injection technique?

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,158 @@
...
+		Configuration zookeeperConfiguration = new ZookeeperConfiguration(zookeeperConnect());
+		ConnectionFactory kafkaConnectionFactory = new DefaultConnectionFactory(zookeeperConfiguration);
+
+		Partition[] partitions = getPartitions();
+
+		KafkaMessageListenerContainer container = ObjectUtils.isEmpty(partitions) ?
+				new KafkaMessageListenerContainer(kafkaConnectionFactory, properties.getTopics()) :
+				new KafkaMessageListenerContainer(kafkaConnectionFactory, partitions);
+
+		container.setOffsetManager(kafkaNativeOffsetManager(kafkaConnectionFactory));
```
artembilan: Oh! I see now. Interesting... :smile: 
Re. `ConnectionFactory` `@Bean`.
Actually the `DefaultConnectionFactory` should really be as a `@Bean`

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
sobychacko: @artembilan Can you please elaborate on this as to what you meant?

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
sobychacko: @artembilan good point. Will change that as a bean

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
artembilan: I mean something like this:

```
@AssertTrue(message = "exactly one of  'topics' and 'partitions' are mutually exclusive")
public boolean isTopicsOrPartitions() {
    return !ObjectUtils.isEmpty(getTopics()) || !ObjectUtils.isEmpty(getPartitions();
}
```

and place it into the `KafkaConfigurationProperties` and here in the `@Configuration` just do `if...else`.

Although I don't know: maybe we should use something as a default topic?..

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,159 @@
...
+
+	private String zkConnectionTimeout = "6000";
+
+	private String zkSessionTimeout = "6000";
+
+	private String zkSyncTime = "2000";
+
+	private String keyDecoder;
+
+	private String payloadDecoder;
```
artembilan: I see that these are converted to the `Class` in the `KafkaSource` thereafter.
The `private Class payloadDecoder;` works well here in the properties.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,159 @@
...
+	private Map<String, String> partitions = new HashMap<>();
+
+	private Map<String, Map<Integer, Long>> initialOffsets = new HashMap<>();
+
+	private String[] topics;
+
+	private int concurrency;
+	private int maxFetch;
+	private int stopTimeout;
+	private int queueSize;
```
artembilan: Blank lines in between class members.
Minor of course, but safe the time during review :smile:

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,159 @@
...
+
+	public Map<String, String> getPartitions() {
+		return partitions;
+	}
+
+	public void setPartitions(Map<String, String> partitions) {
+		this.partitions = partitions;
+	}
+
+	public String getZkConnect() {
```
artembilan: `@NotBlank`

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,159 @@
...
+		return zkConnect;
+	}
+
+	public void setZkConnect(String zkConnect) {
+		this.zkConnect = zkConnect;
+	}
+
+	public String getZkConnectionTimeout() {
```
artembilan: `@NotBlank`

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,159 @@
...
+		return zkConnectionTimeout;
+	}
+
+	public void setZkConnectionTimeout(String zkConnectionTimeout) {
+		this.zkConnectionTimeout = zkConnectionTimeout;
+	}
+
+	public String getZkSessionTimeout() {
```
artembilan: `@NotBlank`

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,158 @@
...
+
+		return container;
+	}
+
+	@Bean
+	public ZookeeperConnect zookeeperConnect() {
```
artembilan: I'd say that `ZookeeperConfiguration` should be as a bean, not this one.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,158 @@
...
+	@Bean
+	public OffsetManager kafkaNativeOffsetManager(ConnectionFactory kafkaConnectionFactory) {
+		return new KafkaNativeOffsetManager(kafkaConnectionFactory, zookeeperConnect(),
+				getInitialOffsets());
+	}
+
+	@Bean
+	public IntegrationFlow fromKafka() throws IllegalAccessException, InstantiationException, ClassNotFoundException {
+
+		return IntegrationFlows.from(Kafka.messageDriverChannelAdapter(container())
```
artembilan: Having such a simple flow, I'd say that there is no reason with the Java DSL overhead.

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceIntegrationTests.java
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * @author Soby Chacko
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {KafkaSourceIntegrationTests.ContextConfiguration.class, KafkaSourceApplication.class},
+		initializers = PropertiesInitializer.class)
+@DirtiesContext
+public abstract class KafkaSourceIntegrationTests {
+
+	private static String scs_kafka_test_embedded;
```
artembilan: I don't like underscore names in Java. Just IMO...
No reason to react, if you think different and leave it as is.

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceIntegrationTests.java
```diff
@@ -0,0 +1,306 @@
...
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
+import static scala.collection.JavaConversions.asScalaBuffer;
+
+/**
+ * Tests for basic {@link KafkaSource}.
```
artembilan, line 66: `basic`? Is there (or planned) an `intermediate` or even `advance`? :smile: 

Just a nit-pick, of course. I can live with this anyway.

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceIntegrationTests.java
```diff
@@ -0,0 +1,306 @@
...
+import org.springframework.integration.kafka.support.KafkaHeaders;
+import org.springframework.integration.kafka.util.TopicUtils;
+import org.springframework.messaging.Message;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
```
artembilan: No one asterisk import.
And some wrong order:
imports

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceIntegrationTests.java
```diff
@@ -0,0 +1,306 @@
...
+		Properties properties = new Properties();
+		properties.put("zkConnect", kafkaTestSupport.getZkConnectString());
+		PropertiesInitializer.PROPERTIES = properties;
+	}
+
+	@AfterClass
+	public static void cleanup() {
+		if (scs_kafka_test_embedded != null) {
+			System.setProperty("SCS_KAFKA_TEST_EMBEDDED", scs_kafka_test_embedded);
+		} else {
```
artembilan: `else` from the new line.
As well as `catch`, `finally` etc..

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaSourceIntegrationTests.java
```diff
@@ -0,0 +1,306 @@
...
+			//Verify that we read from all the partitions and all messages are accounted for
+			Integer[] in = new Integer[]{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11};
+			for (int i : in) {
+				assertTrue(strings.contains(Integer.toString(i)));
+			}
+		}
+	}
+
+	@Configuration
+	@EnableIntegration
```
artembilan: This stuff is redundant for the configuration below.
Just chasing for some test performance. :smile:

pom.xml
```diff
@@ -186,7 +187,7 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-spring-service-connector</artifactId>
 		</dependency>
-		<!--TODO: Move transport related dependencies here pending https://jira.spring.io/browse/XD-3337  -->
+		<!--TODO: Move transporct related dependencies here pending https://jira.spring.io/browse/XD-3337  -->
```
artembilan: M-m-m. The previous one was better. :smile:

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
sobychacko: +1 on the validation stuff. 

I don't think we should default to any topic. We should let the users provide the topic that they want to consume from. If the topic is not present or the topics are not given, then we should simply error out. For the sink case, defining a default topic might make sense.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
sobychacko: @artembilan these are optional values and for zk connect, there is a default value that we fall back to.

artembilan: Well, I mean something like this:

```
@IntegrationTest({"topics = foo", "zkConnect="})
```

when we just can't fall back, because of end-user interaction.

sobychacko:
@artembilan @mbogoevici I tried to address all the comments above. Can you review?

kafka-source/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.stream.module</groupId>
+            <artifactId>spring-cloud-stream-modules-test-support</artifactId>
+            <version>1.0.0.BUILD-SNAPSHOT</version>
```
artembilan: The `version` is redundant here.
We are in the same project with its modules.

I wonder if we even can avoid explicit `1.0.0.BUILD-SNAPSHOT` in the `<parent>` definition.
For example use some external `properties`, like we have with Gradle...

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,181 @@
...
+import javax.validation.constraints.NotNull;
+
+import org.hibernate.validator.constraints.NotBlank;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.ObjectUtils;
+
+/**
+ * Properties related to connecting to Kafka broker.
```
artembilan: I think this JavaDocs isn't relevant. We connect to ZK with the Kafka cluster behind.
Maybe just something like:

```
* The Kafka Source properties.
```

?

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,225 @@
...
+    @Bean
+    public KafkaMessageListenerContainer container() {
+        Partition[] partitions = getPartitions();
+
+        KafkaMessageListenerContainer container = ObjectUtils.isEmpty(partitions) ?
+                new KafkaMessageListenerContainer(kafkaConnectionFactory(), properties.getTopics()) :
+                new KafkaMessageListenerContainer(kafkaConnectionFactory(), partitions);
+
+        // if we have less target partitions than target concurrency, adjust accordingly
+        container.setConcurrency(Math.min(properties.getConcurrency(), partitions.length));
```
artembilan, line 95: This is redundant exercise. See `ConcurrentMessageListenerDispatcher` ctor:

```
this.consumers = Math.min(partitions.size(), consumers);
```

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,225 @@
...
+
+        container.setMaxFetch(properties.getMaxFetch());
+        container.setStopTimeout(properties.getStopTimeout());
+        container.setQueueSize(properties.getQueueSize());
+        container.setOffsetManager(kafkaNativeOffsetManager(kafkaConnectionFactory()));
```
artembilan, line 100: Shouldn't we consider to use the `WindowingOffsetManager` wrapper here as well?
See discussion in the https://github.com/spring-cloud/spring-cloud-stream/pull/316

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationPropertiesTest.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+        catch (Exception e) {
+            assertThat(e, instanceOf(BeanCreationException.class));
+            assertThat(e.getMessage(),
+                    containsString("Either a list of topics (--topics) OR partitions.<topic>=<comma separated partitions> (--partitions.<topic>) must be provided"));
+        }
+    }
+
+    @Test
+    public void testValidTopicSceanrio() throws Exception {
```
artembilan: Typo: Sce**na**rio :smile:

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationPropertiesTest.java
```diff
@@ -0,0 +1,118 @@
...
+        EnvironmentTestUtils.addEnvironment(context, "topics:foo");
+        context.register(Config.class);
+        context.refresh();
+        KafkaConfigurationProperties kafkaConfigurationProperties =
+                context.getBean(KafkaConfigurationProperties.class);
+        assertThat(kafkaConfigurationProperties.getTopics()[0], containsString("foo"));
+    }
+
+    @Test
+    public void testValidPartitionsSceanrio() throws Exception {
```
artembilan: DITO

kafka-source/src/test/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationPropertiesTest.java
```diff
@@ -0,0 +1,118 @@
...
+            AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+            EnvironmentTestUtils.addEnvironment(context, "topics:");
+            context.register(Config.class);
+            context.refresh();
+            fail("BeanCreationException expected");
+        }
+        catch (Exception e) {
+            assertThat(e, instanceOf(BeanCreationException.class));
+            assertThat(e.getMessage(),
+                    containsString("Either a list of topics (--topics) OR partitions.<topic>=<comma separated partitions> (--partitions.<topic>) must be provided"));
```
artembilan, line 54: IMO it's better do not overflow 120 code line length.
At least it is easier to review here on GH.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
sobychacko, line 100: good point. Will wrap it in Windowing..Manager.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
artembilan, line 95: OK. Never mind.
I see now the reason. `ConcurrentMessageListenerDispatcher` has a code like:

```
Assert.isTrue(consumers <= partitions.size(),
                "The number of consumers must be smaller or equal to the number of partitions");
........
this.consumers = Math.min(partitions.size(), consumers);
```

So, we don't reach that `Math.min()` if we provide more consumers than partitions.

I guess that `Assert` should be removed in the SIK. The  `Math.min()` is a good compromise do not disturb end-user.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaConfigurationProperties.java
```diff
@@ -0,0 +1,181 @@
...
+    @NotBlank
+    public String getZkSyncTime() {
+        return zkSyncTime;
+    }
+
+    public void setZkSyncTime(String zkSyncTime) {
+        this.zkSyncTime = zkSyncTime;
+    }
+
+    public Class getKeyDecoder() {
```
artembilan: Should be like `Class<?>` to avoid the compiler noise.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,225 @@
...
+
+        return container;
+    }
+
+    @Bean
+    public KafkaMessageDrivenChannelAdapter kafkaMessageDrivenChannelAdapter() throws IllegalAccessException, InstantiationException, ClassNotFoundException {
```
artembilan: The `ClassNotFoundException` is redundant already.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
```diff
@@ -0,0 +1,225 @@
...
+        final KafkaMessageDrivenChannelAdapter kafkaMessageDrivenChannelAdapter =
```
artembilan, line 107: Since you don't use SI Java DSL anymore, the `spring-integration-java-dsl` dependency can be removed.

kafka-source/src/main/java/org/springframework/cloud/stream/module/kafka/source/KafkaSource.java
mbogoevici, line 95: We can display a warning, I suppose, and label it `maxConcurrency` or something else. The main problem is that they could set a concurrency level that will never be honoured by the running application.

artembilan, line 95: Anyway I think it is a SIK responsibility not here.
Although we should duplicated it here until release there :smile:

mbogoevici:
As part of the ongoing discussion on what types of OOTB modules we want to provide for messaging systems, the Kafka source and sink are too similar to the Kafka binder in terms of functionality, so we will rely on the binder for supporting this functionality (also on multibinder support if needing to connect to different types of systems, or cluster instances). 

Current gaps in the binder: 
- support for multiple topics/destinations in the binder: https://github.com/spring-cloud/spring-cloud-stream/issues/334
- listening to specific partitions: https://github.com/spring-cloud/spring-cloud-stream/issues/336
- resetting to arbitrary offsets: https://github.com/spring-cloud/spring-cloud-stream/issues/335

mbogoevici:
Closed as per above comment.

PR #204, opened 2016-03-02
Make rabbitmq easily customizable

Mwea:


garyrussell:
We typically don't use such prefixes on module-specific properties; otherwise, in SCDF, the deployment would have to specfiy `--properties="spring.rabbitmq.foo=..."` for every property.

This source is a little different from most in that it does use some `spring.rabbitmq.*` properties (broker addresses etc).

But I don't see any compelling reason to add `spring.rabbitmq.` as a prefix to the source-specific properties.

My 2¢

markpollack:
We are in the process of working through this and getting close, modules (not just referred to as apps) will be hosted in the https://github.com/spring-cloud/spring-cloud-stream-app-starters repo and we will be putting a prefix unqiue to each app (and also handling the ability to share common @ConfigurationProperties class) there, so hang on.  In the meantime just use your own fork of the modules.

Issue #191, opened 2016-02-16
Add prefix for OOTB application configs


sabbyanandan:
As a developer, I'd like to add prefix to all the configurations associated with OOTB applications, so we will not run into naming collisions and facilitate persisting the entires in Spring Cloud Config.

**Acceptance:**
- All the OOTB application configurations have prefixes such as log.sink, time.source etc.

ericbottard:
This needs to be discussed further before we implement anything IMHO. Still need to clarify the need, and see how several instances of the same module are expected to be taken care of (*e.g.* the equivalent of `transform --expression=x | transform --expression=y`)

jvalkeal:
Those modules using `port` option complete blows up in CF as it's adding `PORT` env variable.

sabbyanandan:
Depends on spring-cloud/spring-cloud-dataflow#432

sabbyanandan:
spring-cloud/spring-cloud-dataflow#432 generally covers the overall filtering and whitelisting of application properties, so we may not need to explicitly also prefix app names to deploy them in the desired platform.

PR #205, opened 2016-03-02
Make gpfdist easily configurable

Mwea:


jvalkeal:
It's not easy as just adding prefix. All modules need to do this and we need to figure out what actually works, aka #191

markpollack:
We are in the process of working through this and getting close, modules (not just referred to as apps) will be hosted in the https://github.com/spring-cloud/spring-cloud-stream-app-starters repo and we will be putting a prefix unqiue to each app (and also handling the ability to share common @ConfigurationProperties class) there, so hang on.  In the meantime just use your own fork of the modules.

PR #207, opened 2016-03-02
GH-91: RabbitMQ Sink

### LINKED ISSUE #91: Port Rabbit as s-c-s sink
As a s-c-d developer, I'd like to move *rabbit* module from XD to s-c-s repo, so I can use it as *sink* to build streaming pipeline.

Consider addressing [XD-2790](https://jira.spring.io/browse/XD-2790)

**Acceptance:**
* *rabbit* as sink module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *rabbit* module with respective group/artifact/version detail

garyrussell:
Resolves #91

Also addresses https://jira.spring.io/browse/XD-2790

### DIFF
--- a/rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
+++ b/rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
@@ -16,25 +16,24 @@
 
 package org.springframework.cloud.stream.module.rabbit.sink;
 
+import org.springframework.amqp.core.MessageDeliveryMode;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
 import org.springframework.amqp.support.converter.Jackson2JsonMessageConverter;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.cloud.stream.annotation.Bindings;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.config.SpelExpressionConverterConfiguration;
 import org.springframework.cloud.stream.messaging.Sink;
-import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
 import org.springframework.integration.amqp.outbound.AmqpOutboundEndpoint;
 import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.MessageHandler;
-import org.springframework.util.StringUtils;
 
 /**
  * A sink module that sends data to RabbitMQ.
@@ -47,14 +46,10 @@ import org.springframework.util.StringUtils;
 public class RabbitSink {
 
 	@Autowired
-	@Bindings(RabbitSink.class)
-	private Sink channels;
-
-	@Autowired
 	private RabbitSinkProperties properties;
 
-	@Autowired
-	private ApplicationContext context;
+	@Value("#{${converterBeanName:null}}")
+	private MessageConverter messageConverter;
 
 	@ServiceActivator(inputChannel = Sink.INPUT)
 	@Bean
@@ -63,7 +58,8 @@ public class RabbitSink {
 		DefaultAmqpHeaderMapper mapper = new DefaultAmqpHeaderMapper();
 		mapper.setRequestHeaderNames(this.properties.getMappedRequestHeaders());
 		handler.setHeaderMapper(mapper);
-		handler.setDefaultDeliveryMode(this.properties.getDeliveryMode());
+		handler.setDefaultDeliveryMode(this.properties.getPersistentDeliveryMode() ? MessageDeliveryMode.PERSISTENT
+				:MessageDeliveryMode.NON_PERSISTENT);
 		if (this.properties.getExchangeExpression() == null) {
 			handler.setExchangeName(this.properties.getExchange());
 		}
@@ -82,14 +78,8 @@ public class RabbitSink {
 	@Bean
 	public RabbitTemplate rabbitTemplate(ConnectionFactory rabbitConnectionFactory) {
 		RabbitTemplate rabbitTemplate = new RabbitTemplate(rabbitConnectionFactory);
-		String converterBeanName = this.properties.getConverterBeanName();
-		if (StringUtils.hasText(converterBeanName)) {
-			if (RabbitSinkProperties.JSON_CONVERTER.equals(converterBeanName)) {
-				rabbitTemplate.setMessageConverter(jsonConverter());
-			}
-			else {
-				rabbitTemplate.setMessageConverter(this.context.getBean(converterBeanName, MessageConverter.class));
-			}
+		if (this.messageConverter != null) {
+			rabbitTemplate.setMessageConverter(this.messageConverter);
 		}
 		return rabbitTemplate;
 	}
--- a/rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
+++ b/rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.rabbit.sink;
+
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.amqp.support.converter.Jackson2JsonMessageConverter;
+import org.springframework.amqp.support.converter.MessageConverter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.stream.annotation.Bindings;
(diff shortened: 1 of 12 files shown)


rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * A SpEL expression that evaluates to a routing key.
+	 */
+	private Expression routingKeyExpression;
+
+	/**
+	 * Default delivery mode when 'amqp_deliveryMode' header is not present.
+	 */
+	private MessageDeliveryMode deliveryMode = MessageDeliveryMode.NON_PERSISTENT;
```
artembilan: Can't we simplify it into something like `boolean persistentDeliveryMode;`?
It would be much easier to type the module options from the console.

From other side looks like it isn't consistent with the default one in the `MessageProperties`:

```
static final MessageDeliveryMode DEFAULT_DELIVERY_MODE = MessageDeliveryMode.PERSISTENT;
```

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Headers that will be mapped.
+	 */
+	private String mappedRequestHeaders[] = { "*" };
```
artembilan, line 63: Not sure in the problem with that https://jira.spring.io/browse/XD-2790. It is `sink`, not `binder`, so that is normal that end-user application transfers into the RabbitMQ just only standard header like it is done by default:

```
public DefaultAmqpHeaderMapper() {
    super(AmqpHeaders.PREFIX, STANDARD_HEADER_NAMES, STANDARD_HEADER_NAMES);
}
```

Otherwise yes, you can configure any desired headers via `mappedRequestHeaders`.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * The bean name for a custom message converter; if omitted, a SimpleMessageConverter is used.
+	 * If 'jsonConverter', a Jackson2JsonMessageConverter bean will be created; for other bean
+	 * names, you must provide the bean in the context.
```
artembilan: Wouldn't this description be so long for the console when we display the module options?

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
```diff
@@ -0,0 +1,103 @@
...
+ * @author Gary Russell
+ */
+@EnableBinding(Sink.class)
+@EnableConfigurationProperties(RabbitSinkProperties.class)
+@Import(SpelExpressionConverterConfiguration.class)
+public class RabbitSink {
+
+	@Autowired
+	@Bindings(RabbitSink.class)
+	private Sink channels;
```
artembilan: This variable is out of use and I wonder if we really need it here...
Seems for me the `@EnableBinding(Sink.class)` is enough. 
Even if we would have several Sink in the same classpath, this `@Autowired` won't help to that `@ServiceActivator` definition.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
```diff
@@ -0,0 +1,103 @@
...
+	@Bean
+	public RabbitTemplate rabbitTemplate(ConnectionFactory rabbitConnectionFactory) {
+		RabbitTemplate rabbitTemplate = new RabbitTemplate(rabbitConnectionFactory);
+		String converterBeanName = this.properties.getConverterBeanName();
+		if (StringUtils.hasText(converterBeanName)) {
+			if (RabbitSinkProperties.JSON_CONVERTER.equals(converterBeanName)) {
+				rabbitTemplate.setMessageConverter(jsonConverter());
+			}
+			else {
+				rabbitTemplate.setMessageConverter(this.context.getBean(converterBeanName, MessageConverter.class));
```
artembilan: I found that this looks better:

```
@Value("#{${converterBeanName:null}}")
private MessageConverter messageConverter;
```

Much closer to that what we can have in the XML config. :smile:

spring-cloud-stream-modules-docs/src/main/asciidoc/sinks.adoc
```diff
@@ -514,11 +514,36 @@ and you should see the following output in the Spring Cloud Data Flow console.
...
+$$exchangeExpression$$:: $$a SpEL expression that evaluates to the Exchange on the RabbitMQ broker to which messages
+should be sent; overrides `exchange`$$ *($$String$$, default: ``)*
+$$mappedRequestHeaders$$:: $$request message header names to be propagated to RabbitMQ, to limit to the set of standard headers plus `bar`, use `STANDARD_REQUEST_HEADERS,bar`$$ *($$String$$, default: `*`)*
+$$routingKey$$:: $$the routing key to be passed with the message, as a SpEL expression$$ *($$String$$, default: none)*
+$$routingKeyExpression$$:: $$an expression that evaluates to the routing key to be passed with the message, as a SpEL expression; overrides `routingKey`$$ *($$String$$, default: none)*
+
+NOTE: By default, the message converter is a `SimpleMessageConverter` which handles `byte[]`, `String` and
+`java.io.Serializable`.
+A well-known bean name `jsonConverter` will configure a `Jackson2JsonMessageConverter` instead.
+In addition, a custom converter bean can be added to the context and referenced by this property.
```
artembilan: `... and referenced by the  $$converterBeanName$$ property.`
Otherwise it isn't clear what is `this` here.

spring-cloud-stream-modules-docs/src/main/asciidoc/sinks.adoc
```diff
@@ -514,11 +514,36 @@ and you should see the following output in the Spring Cloud Data Flow console.
...
 
 [[spring-cloud-stream-modules-redis]]
 === Redis (`redis`)
 
-Redis sink can be used to ingest data into redis store. You can choose `queue`, `topic` or `key` with selcted collection type to point to a specific data store.
+The Redis sink can be used to ingest data into redis store. You can choose `queue`, `topic` or `key` with selcted
+git ccollection type to point to a specific data store.
```
artembilan: Typo: double `c`.
And what does `git` do here? :smile:

artembilan:
That's all from my side.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
garyrussell: I decided to make it non-persistent by default here for performance reasons.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
garyrussell, line 63: I don't understand your point; the binder already uses `*` as the default; the JIRA asks for the same for the source and sink.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
artembilan: OK. That's fine, too. Just wanted to know the premise. 
How about the `boolean` style, BTW ?

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
artembilan, line 63: I agree that binder should definitely do that as a `*`, but from other side that JIRA doesn't explain the reason for Source and Sink.
Just a little nit-picking. Maybe that should be pushed to SI directly, because the default there is a bit different...

CC: @dturanski

dturanski, line 63: Hmmm. I don't remember the specific motivation for the JIRA. But if setting headers in a stream, I would expect them to be propagated by default.

artembilan, line 63: OK. We can live with this as well. Although it seems for me much easier to specify `*` from the console than the long constant for standard headers to restrict.
I wonder how much non-standard headers SCDF adds...

garyrussell, line 63: It's not so much what SCDF adds it's what users add.

Users complained that with `...->headerEnricher->rabbit`  the headers are stripped (by default).

TBH, I don't really care either way; however, I **have** struggled a bit with the inconsistencies in SI (JMS maps all headers, Rabbit doesn't).

artembilan, line 63: Well, with the `negated patterns` I'm sure we can change the default for the AMQP header mapping to the all since SI-4.3, like it is with JMS.
And may review all others which supports `negated patterns`, too...

artembilan:
I'm fine to merge it, but there are several unanswered comments, which I won't do on merge silently.

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSinkProperties.java
garyrussell: I don't know how to make it shorter.

garyrussell:
Pushed

rabbit-sink/src/main/java/org/springframework/cloud/stream/module/rabbit/sink/RabbitSink.java
```diff
@@ -0,0 +1,93 @@
...
+
+	@ServiceActivator(inputChannel = Sink.INPUT)
+	@Bean
+	public MessageHandler amqpChannelAdapter(ConnectionFactory rabbitConnectionFactory) {
+		AmqpOutboundEndpoint handler = new AmqpOutboundEndpoint(rabbitTemplate(rabbitConnectionFactory));
+		DefaultAmqpHeaderMapper mapper = new DefaultAmqpHeaderMapper();
+		mapper.setRequestHeaderNames(this.properties.getMappedRequestHeaders());
+		handler.setHeaderMapper(mapper);
+		handler.setDefaultDeliveryMode(this.properties.getPersistentDeliveryMode() ? MessageDeliveryMode.PERSISTENT
+				:MessageDeliveryMode.NON_PERSISTENT);
```
artembilan, line 62: Whitespace after colon.
Where is our friend Checkstyle? :smile: 

Will fix on merge.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/4b9b5779ee8bdee9856d262797b70b03b5bac6de.

Any thoughts about the default `*` for AMQP adapters in SI?

garyrussell:
Sure: https://jira.spring.io/browse/AMQP-588

PR #208, opened 2016-03-03
[DO NOT MERGE] Add dataflow.properties file to modules

ericbottard:
This is a placeholder PR for when https://github.com/spring-cloud/spring-cloud-dataflow/pull/432 is in.

At the moment, only the `cassandra-sink` module is concerned, to ease review of the aforementioned PR.


markfisher:
Is this really the only way to support filtering of the properties for completion?

I am strongly opposed to anything that requires a `spring-cloud-stream` app developer to be aware of Data Flow. cc @mbogoevici

ericbottard:
Closing this for now, as we slightly change approach

PR #209, opened 2016-03-03
GH-67: Add Syslog Source

### LINKED ISSUE #67: Port syslog as s-c-s source
As a developer, I'd like to move *syslog* module from XD to s-c-s repo, so I can use it as source to build streaming pipeline. 
**Acceptance:**
* *syslog* as source module is available in *spring-cloud-stream-modules* repo 
* maven repo includes *syslog* module with respective group/artifact/version detail

garyrussell:
Resolves #67

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
 		<module>loggregator-source</module>
 		<module>rabbit-source</module>
 		<module>sftp-source</module>
+		<module>syslog-source</module>
 		<module>tcp-source</module>
 		<module>time-source</module>
 		<module>trigger-source</module>
--- a/spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
+++ b/spring-cloud-stream-modules-docs/src/main/asciidoc/sources.adoc
@@ -216,7 +216,25 @@ $$tmpFileSuffix$$:: $$extension to use when downloading files$$ *($$String$$, de
 $$user$$:: $$the username to use$$ *($$String$$, no default)*
 $$withMarkers$$:: $$if true emits start of file/end of file marker messages before/after the data. Only valid with FileReadingMode 'lines'$$ *($$Boolean$$, no default)*
 
-[[spring-cloud-stream-modules-source-tcp]]
+[[spring-clound-stream-modules-source-syslog]]
+=== SYSLOG
+
+The syslog source receives SYSLOG packets over UDP, TCP, or both.
+RFC3164 (BSD) and RFC5424 formats are supported.
+
+==== Options
+
+The **$$syslog$$** $$source$$ has the following options:
+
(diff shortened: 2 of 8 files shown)


jvalkeal:
Btw, just one more module which have zero change to work on PCF due to non-prefix `port` option which breaks with PCF `PORT` env variable.

markpollack:
A prefix will be added as part of the great app migration of 2016 to spring-cloud-stream-app-starters.

markpollack:
I'm getting test failures on `NotNioTests`, `PropertiesPopulatedTests`, `Tcp3164Tests` and `Tcp5424Tests`.  

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'tcpAdapter' defined in class path resource [org/springframework/cloud/stream/module/syslog/source/SyslogSource.class]: Unsatisfied dependency expressed through constructor argument with index 0 of type [org.springframework.integration.ip.tcp.connection.AbstractServerConnectionFactory]: No qualifying bean of type [org.springframework.integration.ip.tcp.connection.AbstractServerConnectionFactory] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {@org.springframework.beans.factory.annotation.Qualifier(value=syslogSourceConnectionFactory)}; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [org.springframework.integration.ip.tcp.connection.AbstractServerConnectionFactory] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {@org.springframework.beans.factory.annotation.Qualifier(value=syslogSourceConnectionFactory)}
```

Not sure why the factory bean `TcpConnectionFactoryFactoryBean` isn't kicking in just for these tests, as the others pass.

Looks like `TcpConnectionFactoryFactoryBean extends AbstractFactoryBean<AbstractConnectionFactory>` and not `AbstractServerConnectionFactory`

markpollack:
I got the tests to pass if I changed the @Bean defs method args from `AbstractServerConnectionFactory` to `AbstractConnectionFactory` but then had to downcast in 

```
    private TcpSyslogReceivingChannelAdapter createTcpAdapter(AbstractConnectionFactory connectionFactory) {
        TcpSyslogReceivingChannelAdapter adapter = new TcpSyslogReceivingChannelAdapter();
        adapter.setConnectionFactory((AbstractServerConnectionFactory)connectionFactory);
        setAdapterProperties(adapter);
        return adapter;
    }
```

which feels icky, even if there was an instance of check...

garyrussell:
Strange - I am not getting any failures - either in STS or maven - but it's likely due to early access to the FB getObjectType() - I'll get rid of the FB.

garyrussell:
Rebased, pushed.

markpollack:
squashed and merged as e26b0a006f7ca8431f7f27b1288617436141483f

PR #217, opened 2016-03-12
GH-216: Rabbit Source - Don't Map DeliveryMode

### LINKED ISSUE #216: Rabbit Source: Don't Propagate Delivery Mode
See https://github.com/spring-projects/spring-integration/pull/1726 and https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java#L199

Currently the inbound delivery mode will override the binder setting when sending; the header should be stripped - Spring Integration 4.3 will strip it by default.

garyrussell:
Resolves #216

Suppress the delivery mode header to avoid inadvertent propagation when
using the rabbit binder.

Also revert the default inbound mapping to only `STANDARD_REQUEST_HEADERS` since
unexpected results may occur when mapping all headers.

See the CAUTION here: http://docs.spring.io/spring-integration/reference/html/amqp.html#amqp-message-headers


artembilan:
Nothing to complain about.
Merging...

Issue #200, opened 2016-02-29
Add README for all OOTB apps


sabbyanandan:
As a developer, I'd like to add README for all OOTB stream applications, so we have a consistent approach for reusing docs both at each application level and in the ref. guide.

Issue #193, opened 2016-02-16
Add a README for every module


dsyer:
I can see quite a few with no, or inadequate, descriptions. Ideally I'd like to know how to use each one as a standalone app, plus details (examples) of configuration options, and mandatory configuration (e.g. credentials).

sabbyanandan:
@dsyer: We wanted not to duplicate docs, so currently all the docs are in the [reference guide](http://docs.spring.io/spring-cloud-stream-modules/docs/current-SNAPSHOT/reference/html/) itself. That said, we could certainly link the guide in all the READMEs.

dsyer:
Could it work the other way (include all the module READMEs in the main guide)?

sabbyanandan:
The application specific content is categorized into [source, processor, and sink ascii docs](https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/spring-cloud-stream-modules-docs%2Fsrc%2Fmain%2Fasciidoc%2Findex.adoc). These docs are not just used in SCSM, but also in each SCDF SPI reference guide as includes, too. We will have to explore how to automate it, so there's always one source for docs across all the different guides/readme's. 

We will look into it.

markfisher:
in the same theme of making these modules into reusable libraries, I would prefer the docs be within each module project, then making the high-level doc a derived "view" of those

ericbottard:
I'm also in favor of pulling the READMEs (or sections thereof) _into_ the reference guide, not the other way around

Issue #153, opened 2016-01-04
Automate module doc updates to ref. guide


sabbyanandan:
As a developer, I'd like to automate any updates to module docs to reference guide, so I don't have to manually maintain them.

This is related to #193

PR #222, opened 2016-03-16
Use READMEs for reference doc

ericbottard:
### Update

This PR does the following: clean up the asciidoctor configuration to a best extent (this was obviously copy/pasted from boot and some bits did not apply).

Also references the individual READMEs of each application module, using the [partial include](http://asciidoctor.org/docs/user-manual/#include-partial) feature.

Also, all READMEs have been proofread, changing/removing references to dataflow, as well as the word "module" to the best extent.

I suggest we merge this, and then we'll port to the starters repo.
### Original PR text

Fixes spring-cloud/spring-cloud-stream-modules#200
Fixes spring-cloud/spring-cloud-stream-modules#193
Fixes spring-cloud/spring-cloud-stream-modules#237
Fixes spring-cloud/spring-cloud-stream-modules#153

This is the simple template I propose we use for READMEs. Of note:
- the options section could be generated in the future, either via an asciidoc extension (would require new version of the maven asciidoctor plugin) or plain old text generation as we had in XD.
- The title drops "Spring Cloud Stream" as it is intended that those READMEs will be `include`d in the reference documentation of Spring Cloud Stream modules.

Let's discuss here what we want to add to the template before we generalize.


markpollack:
I think we should keep the simple snippits of how to build and run.  It is some boilerplate but doesn't hurt.  We can remove the existing list of classes and requirements.

markpollack:
We have a different type of http source now, based of spring mvc - what does this mean for all the https functionality listed here in XD http://docs.spring.io/spring-xd/docs/current/reference/html/#http ?

ericbottard:
> I think we should keep the simple snippits of how to build and run. It is some boilerplate but doesn't hurt. 

Gotcha. Should we still include those (and hence repeat them _ad nauseam_) in the reference doc though? Asciidoctor supports conditional includes (but not excludes) of zones, so would be possible to exclude them, although a bit more setup.

ericbottard:
> We have a different type of http source now, based of spring mvc - what does this mean for all the https functionality listed here in XD

I think we agreed at some point that we'd revisit the http source anyway, using SI components. cc @markfisher @garyrussell

markpollack:
Yes I believe so.  Seems a small price to pay for someone new coming to the
page without too much overall context.  I do see how it isn't useful in the context of DataFlow to just do a raw include.  Can I merge this (adding the 'how to run' bits) and then another story for investigating conditional includes?

ericbottard:
> Can I merge this (adding the 'how to run' bits) and then another story for investigating conditional includes?

Sorry only seeing this last sentence now. I'd rather continue working on the whole README refactoring, and we can merge all at once, wdyt ?

jvalkeal:
Documenting things like this where only non-prefix options are written in readme shows the trouble from user perspective. Like http source only have one option, `pathPattern`. Order to use it you need to set a port(defaults to random port) which is done via `server.port`. Normal user would not know this as it's not in a docs. Listing all options in shell is still pointless as there's too many stuff and you never know which ones are meaningful for a particular module.

ericbottard:
@markpollack This PR is now ready for review, I updated the description.

markpollack:
Yes, there is to be a white list of properties that are appropriate to use which come directly from boot.  The docs should include those as well.

Each app will be getting its own unique prefix for its own options.  Completion in the shell will be aware of th app prefix conventions and the white list.

The docs will be updated to include the app prefix once it has been merged in.  Eric is working on this.

markpollack:
Will deal with the whitelisting as part of another PR.

markpollack:
rebased and fixed merge conflicts as 67cc61cf675477d2a8ef899b27abe48f7353ad09

PR #227, opened 2016-03-17
Scriptable transform processor

aclement:
First drop of the transform processor that can be configured by passing a script directly. Supports groovy/javascript/ruby. The README includes examples of using it from the shell but obviously the UI provides a nicer script editing experience and also looks after the necessary escaping/unescaping when producing the DSL string.

Open questions:
- is 'script' the right property name or will users expect that to take a filename like it does in other modules? If so, what is a better name?
- I have temporarily removed python support because it worked when run standalone but failed under SCDF and I can't easily debug the module under SCDF right now.
- Should this module have an additional parameter that allows you to pass in a script file reference? I did add that but later removed it when I decided I wanted this module to concentrate on one thing.


artembilan:
A couple general comments for now:
1. We can absorb existing `groovy-transformer-processor`
2. Take a look, please, into the SI Java DSL Scripting support: https://github.com/spring-projects/spring-integration-java-dsl/blob/master/src/test/java/org/springframework/integration/dsl/test/scripts/ScriptsTests.java. Maybe that would be better to utilize it from there.
3. Anyway it would be better to bypass `ScriptExecutorFactory` for Groovy. That `script engine` is definitely very slow: https://gist.github.com/artembilan/7072431. We should use `GroovyScriptExecutingMessageProcessor`. That is included to the SI Java DSL Scripting support.

aclement:
I applied the suggestion form Artem and switched to the Spring Integration Scripts class. This includes that additional benefit of a more optimal groovy script execution path.

>  We can absorb existing groovy-transformer-processor

This calls back to my third question. We can only replace groovy-transform-processor if this module is further enhanced to also support a script-path type property too. I haven't had any feedback on that. I don't mind doing it but it would mean more property validation (have they supplied this property or that property, ok well if they supplied that one then I need to check they specified a language, blah blah blah).

scriptable-transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/ScriptableTransformProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+	@Transformer(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
+	public MessageProcessor<?> transformer() {
+		String language = properties.getLanguage();
+		String script = properties.getScript();
+		logger.info("Input script is '{}', language is '{}'", script, language);
+		Resource scriptResource = new ByteArrayResource(decodeScript(script).getBytes()) {
+			@Override
+			public String getFilename() {
+				// Only the groovy script processor enforces this requirement for a name
+				return "StaticScript";
```
artembilan, line 73: I've raised SI JIRA on the matter: https://jira.spring.io/browse/INT-3976

scriptable-transform-processor/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+			<groupId>org.jruby</groupId>
+			<artifactId>jruby-complete</artifactId>
+			<version>9.0.5.0</version>
+		</dependency>
+		<!-- 
+		<dependency>
+			<groupId>org.python</groupId>
+			<artifactId>jython-standalone</artifactId>
+			<version>2.7.0</version>
+		</dependency>
```
artembilan, line 36: Actually it doesn't care us that SCDF doesn't work (yet?) with Python properly.
We are SCSt here and this application definitely can be run as a standalone outside of SCDF.
If all test passes, I'll uncomment the Python support on merge.

artembilan:
A general comment: should we still add content to the documentation alongside with the `README`?
I mean the note for this processor in the `processors.adoc`.

CC @sabbyanandan

sabbyanandan:
@artembilan: For now, yes. We are yet to automate the auto-population of README into the reference guide. So, let's please do it in both places.

artembilan:
Thank you @sabbyanandan !

Will do on merge, @aclement

scriptable-transform-processor/README.adoc
```diff
@@ -0,0 +1,56 @@
...
+```
+
+```
+XD> stream create --deploy true --name demojs --definition "time | scriptable --script=\"function double(a) {\\n return a+''+a;\\n}\\ndouble(payload);\" --language=js | log"
+```
+
+Python is not currently working but when it does it will look like this!
+
+```
+XD> stream create --deploy true --name demop --definition "time | scriptable --script=\"def concat(x,y):\\n  return x+y\\nanswer = concat(\"\"hello \"\",payload)\\n\" --language=python | log"
```
artembilan, line 55: No any `XD` word in the SCSt :smile: 

Although I guess it would be better do not position our code samples here on the `dataflow` as well.
But the cat is already out of the bag!

So, I'll just fix this for `dataflow` as it is in any other place here...

scriptable-transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/ScriptableTransformProcessorProperties.java
```diff
@@ -0,0 +1,72 @@
...
+ * @author Andy Clement
+ */
+@ConfigurationProperties
+public class ScriptableTransformProcessorProperties {
+
+	/**
+	 * Language of the text in the script property. Supported: groovy, javascript, ruby.
+	 */
+	@NotNull
+	private String language;
```
artembilan, line 34: Since it is required option (yet), how about to make it shorter - `lang`?

artembilan:
Polishing for confirmation before merging: https://github.com/artembilan/spring-cloud-stream-modules/commit/4c3da02a21cdab97ff002225367675a26040859c

scriptable-transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/ScriptableTransformProcessorProperties.java
aclement, line 34: sure thing. I am also still a bit nervous that the other option is called 'script' as anyone with XD experience may think that takes a resource path/url rather than code directly (since that is what script does in other modules).

artembilan:
> is 'script' the right property name or will users expect that to take a filename like it does in other modules? If so, what is a better name?

I think that's fine. when we introduce a file support it would be like `scriptLocation`. Pay attention we already have there `variablesLocation`.

scriptable-transform-processor/src/main/java/org/springframework/cloud/stream/module/transform/ScriptableTransformProcessorProperties.java
artembilan, line 34: From one hand he must go to the RTFM do not show himself a bit silly.
From other side it is an argument to introduce a `scriptLocation` ASAP :smile:

artembilan:
> I have temporarily removed python support because it worked when run standalone but failed under SCDF and I can't easily debug the module under SCDF right now.

I have never ran the Dataflow, so can't confirm how that doesn't work. But I'm sure that is out of SCSt scope. See my polishing: I uncommented the Python stuff.

artembilan:
> Should this module have an additional parameter that allows you to pass in a script file reference?

We can do that eventually. Just need to agree within the team and raise a story on the matter, if that.

artembilan:
That's all.
Let me know you opinion about my polishing and I even can change the property just to the `lang` on merge.

aclement:
Polishing looks good. Sorry I didn't have a decent set of formatting options installed that match the current expectations in the codebase. Your change for lang would be good and to keep script and add scriptLocation later all sound fine to me. Thanks Artem.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/95c8774ff27768e25c9a1fc4558e3f82a5202850
Also added: https://github.com/spring-cloud/spring-cloud-dataflow/commit/400a512bab2f4cc16286a96f622cf23bc73891bc

Thank you @aclement for as always cool and very smart work!

Issue #214, opened 2016-03-10
Make `redis` as the default store for all the counters


sabbyanandan:
As a user, I'd like to use `redis` instead of `in-memory` as the default store for `field-value-counter`, `counter`, and `aggregate-counter`.

**Acceptance:**
- `field-value-counter`, `counter` and `aggregate-counter` (_when spring-cloud/spring-cloud-stream-modules#199 is merged_) by default uses `redis` as the store
- `--store` option defaults to `redis` in the docs

PR #230, opened 2016-03-18
Use redis as the default store for analytics

ericbottard:
Fixes spring-cloud/spring-cloud-stream-modules#214

No tests were harmed during the making of this PR.


markfisher:
LGTM! merging

Issue #233, opened 2016-03-21
Make integration-test-processor a plain boot app


ericbottard:
Also, the thinner this app gets, the better, as the CF v2 API currently does not support differential upload.

PR #234, opened 2016-03-23
Simplify app so that it does not depend on scs.

ericbottard:
Fixes spring-cloud/spring-cloud-stream-modules#233

### DIFF
--- a/integration-test-processor/pom.xml
+++ b/integration-test-processor/pom.xml
@@ -1,19 +1,48 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
+	<groupId>org.springframework.cloud.stream.module</groupId>
 	<artifactId>integration-test-processor</artifactId>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
 	<packaging>jar</packaging>
 	<description>Spring Cloud Stream module used to perform integration tests</description>
 
 	<parent>
-		<groupId>org.springframework.cloud.stream.module</groupId>
-		<artifactId>spring-cloud-stream-modules</artifactId>
-		<version>1.0.0.BUILD-SNAPSHOT</version>
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>1.3.3.RELEASE</version>
 	</parent>
 
 	<properties>
 		<start-class>org.springframework.cloud.stream.module.it.IntegrationTestProcessorApplication</start-class>
+		<java.version>1.8</java.version>
 	</properties>
-	
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-configuration-processor</artifactId>
+			<!--optional so that it does its job while compiling, but is not included in
+			the resulting artifact -->
+			<optional>true</optional>
+		</dependency>
+	</dependencies>
+	<build>
+		<plugins>
+			<plugin>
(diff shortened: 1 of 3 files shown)


integration-test-processor/pom.xml
```diff
@@ -1,19 +1,48 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
+	<groupId>org.springframework.cloud.stream.module</groupId>
```
markfisher, line 6: should we change this groupId?... if it exists solely for the deployer SPI TCK, this could be "org.springframework.cloud.deployer"

integration-test-processor/pom.xml
```diff
@@ -1,19 +1,48 @@
...
+			<optional>true</optional>
+		</dependency>
+	</dependencies>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+				<configuration>
+					<classifier>exec</classifier>
```
markfisher, line 42: do we need the classifier?

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
```diff
@@ -22,25 +22,18 @@
...
-import org.springframework.cloud.stream.messaging.Processor;
 
 /**
  * A module that can misbehave, useful for integration testing of module deployers.
  *
  * @author Eric Bottard
  */
-@EnableBinding(Processor.class)
 @EnableConfigurationProperties(IntegrationTestProcessorProperties.class)
 public class IntegrationTestProcessor {
```
markfisher, line 32: this is not really a "Processor" is it?

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessorProperties.java
```diff
@@ -56,6 +57,12 @@
 	 */
 	private String parameterThatMayNeedEscaping;
 
+	/*
+	 * Same as ChannelBindingServiceProperties.instanceIndex.
```
markfisher, line 61: why mention "ChannelBindingServiceProperties"  here?

integration-test-processor/pom.xml
ericbottard, line 6: We could even move this ~~module~~ app out of this repo and put it next to the new deployer code (and switch groupId) once all dataflow implementations use the new deployer code.
It is currently used by the (old) CF deployer implementation TCK for example.

integration-test-processor/pom.xml
ericbottard, line 42: I kept it for consistency with other modules, but the more we make this standalone, the less it needs to look like another module. Would need to change the TCK abstract class(es) though.

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessor.java
ericbottard, line 32: Definitely not anymore, no. Note that no test exercised its processor nature before anyway.

integration-test-processor/pom.xml
mbogoevici, line 6: I believe it would make much more sense if this module lived somewhere else altogether.

integration-test-processor/pom.xml
markfisher, line 42: since this is for the deployer SPI, I don't think the conventions used for those particular apps are relevant

integration-test-processor/src/main/java/org/springframework/cloud/stream/module/it/IntegrationTestProcessorProperties.java
ericbottard, line 61: Because this now mimics what is handled by `ChannelBindingServiceProperties.instanceIndex`. This is intended for a test that deploys multiple instances of the same app. But I realize that test never made it to `AbstractAppDeployerIntegrationTests` o.O ?

markfisher, line 61: it doesn't matter if it's the same as `ChannelBindingServiceProperties.instanceIndex` - that's inverting the relationship... everything about this app should be as general as possible - anything related to `spring-cloud-stream` apps is more specific than this

ericbottard:
Closing, working on a PR against https://github.com/spring-cloud/spring-cloud-deployer

PR #235, opened 2016-03-23
removing lombok from web socket sink and avoiding 9393 port conflict in test

markfisher:


mbogoevici:
While we're at it, could we just rename this module altogether? While it is valuable to have Netty in, it feels like the default websocket implementation could be a much simpler approach.

markfisher:
+1 for creating a new websocket sink altogether, and qualifying this one

garyrussell:
LGTM, except I think it would be even better to use a random port for the test.

Polishing here for your approval b4 merging... https://github.com/garyrussell/spring-cloud-stream-modules/commit/66a562b3d67546cb338b245d5205f89256039784

markfisher:
Yes, the random port is much better. Thanks!

garyrussell:
rebased, merged as bb6f1bfc0aa5a938fa48faf3c3ebac1491e8c303

PR #243, opened 2016-03-30
Refactor `lang` -> `language`

artembilan:
The `lang` name for application property isn't good enough and may clash with ENV var or System property.
Change the property to the `language` back.

The future story with the global common prefix for apps should fix similar confuse.


dturanski:
LGTM

Issue #245, opened 2016-03-31
GemfireSink must declare CacheWritingMessageHandler as a bean


artembilan:
Since `GemfireSink` doesn't declare `CacheWritingMessageHandler` as a bean we end up with the `NPE` for the `eveluationContext` at runtime:
```
Caused by: java.lang.IllegalArgumentException: EvaluationContext is required
	at org.springframework.util.Assert.notNull(Assert.java:115) ~[spring-core-4.2.5.RELEASE.jar!/:4.2.5.RELEASE]
	at org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:249) ~[spring-expression-4.2.5.RELEASE.jar!/:4.2.5.RELEASE]
	at org.springframework.integration.gemfire.outbound.CacheWritingMessageHandler.evaluateCacheEntries(CacheWritingMessageHandler.java:110) ~[spring-integration-gemfire-4.2.4.RELEASE.jar!/:na]
	at org.springframework.integration.gemfire.outbound.CacheWritingMessageHandler.handleMessageInternal(CacheWritingMessageHandler.java:82) ~[spring-integration-gemfire-4.2.4.RELEASE.jar!/:na]
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:127) ~[spring-integration-core-4.2.4.RELEASE.jar!/:na]
	... 35 common frames omitted
```
The fix is pretty simple:
````java
@EnableBinding(Sink.class)
@Import({GemfirePoolConfiguration.class, GemfireClientRegionConfiguration.class})
@EnableConfigurationProperties(GemfireSinkProperties.class)
public class GemfireSink {
	
	@Autowired
	GemfireSinkProperties config;

	@Resource(name = "clientRegion")
	Region<String,?> region;
	
	@Bean
	@ServiceActivator(inputChannel = Sink.INPUT)
	public MessageHandler cacheWritingMessageHandler() {
		CacheWritingMessageHandler messageHandler = new CacheWritingMessageHandler(region);
		messageHandler.setCacheEntries(Collections.singletonMap(config.getKeyExpression(), "payload"));
		return messageHandler;
	}
	
}
````

PR #246, opened 2016-04-01
Resolves https://github.com/spring-cloud/spring-cloud-stream-modules/issues/245

dturanski:
and probably https://github.com/spring-cloud/spring-cloud-stream-modules/issues/49

### DIFF
--- a/gemfire-sink/src/test/java/org/springframework/cloud/stream/module/gemfire/GemfireSinkIntegrationTests.java
+++ b/gemfire-sink/src/test/java/org/springframework/cloud/stream/module/gemfire/GemfireSinkIntegrationTests.java
@@ -100,8 +100,6 @@ public class GemfireSinkIntegrationTests {
 	}
 
 	@Test
-	@Ignore("Need to spawn an embedded gemfire cacahe server.  " +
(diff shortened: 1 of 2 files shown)


artembilan:
I'll take care of this.

artembilan:
Ha! @sabbyanandan has already beat me on that :smile:

gemfire-sink/src/test/java/org/springframework/cloud/stream/module/gemfire/GemfireSinkIntegrationTests.java
```diff
@@ -100,8 +100,6 @@ public boolean waiting() {
 	}
 
 	@Test
-	@Ignore("Need to spawn an embedded gemfire cacahe server.  " +
```
artembilan, line 103: Well, we haven't caught the `NPE` before because this was an only test-case for `GemfireSink` and it was ignored...
:cry:

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/d27c17bc84e89e2d68016d3cd575de0aceed387c

PR #249, opened 2016-04-06
Fix counter sink

ilayaperumalg:
- Fix logic that creates RedisMetricRepository
  - Add test to assert redis metric repository

This resolves #248

PR #250, opened 2016-04-12
Use correct store in counter config

ractive:


ilayaperumalg:
Hi @ractive,

Thank you for the PR. This has already been addressed in #249.
Also, we are in the process of moving the apps into `spring-cloud-stream-app-starters` repo [here](https://github.com/spring-cloud/spring-cloud-stream-app-starters) and hence the actual fix was made there.

Hence, I am closing this PR. Thank you!

PR #251, opened 2016-04-13
Add `jsonpath` as dependency

sabbyanandan:
Without the dependency, when using `jsonpath` in `--expression` property, I get the following error.

```
2016-04-13T12:19:33.29-0700 [APP/0]      OUT Caused by: org.springframework.expression.spel.SpelEvaluationException: EL1022E:(pos 0): The function 'jsonPath' mapped to an object of type 'class org.springframework.expression.TypedValue' which cannot be invoked
```

With this change, I can interact with the payload as: 

```
stream create --name filterTest --definition "http | transform --expression=\"#jsonPath(payload,'$._output')\" | log" --deploy
```

This is applicable for any application that supports `--expression` property; hence, it is included in the root `pom`.

_note: this change needs ported over to app-starter eventually; the assembly mechanism is different there since we do not have a root-level `pom`_

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,6 +183,11 @@
 				<scope>test</scope>
 			</dependency>
 			<dependency>
+				<groupId>com.jayway.jsonpath</groupId>
+				<artifactId>json-path</artifactId>
+				<version>2.2.0</version>
+			</dependency>
+			<dependency>
 				<groupId>org.springframework.cloud.stream.module</groupId>
 				<artifactId>script-variable-generator</artifactId>
 				<version>1.0.0.BUILD-SNAPSHOT</version>


artembilan:
@sabbyanandan , 

should we really apply this change to the `modules`, since the workaround is enough simple with an additional dependency at runtime?
Maybe we should just document the IMPORTANT note here?..

I tried to test it locally in the `transform-processor`:

``` java
@Autowired
@Qualifier("&jsonPath")
private SpelFunctionFactoryBean jsonPathSpelFunctionFactoryBean;
```

Without success, because `json-path` dep must be declared explicitly in the modules pom, on in the root pom, but already in the `<dependencies>` block not only in this `<dependencyManagement>`.

From one side it looks better to specify such a dep only in those modules which may rely on the SpEL, like this `transformer` or `splitter`. But from other hand we can't predict which end-user application may expose similar SpEL behavior, so `#jsonPath()` would be good there as well. So, from here it might be better to always provide the `json-path` dependency and allow picky users to exclude it in his environment.

Plus, as we talked privately, the `spring-integration-xml` with its built-in `#xpath()` should be there as well for the same SpEL reason.

So, my position is like:
1. Do nothing here in modules, unless document.
2. apply `json-path` and `spring-integration-xml` deps in the `apps` as top-level for all applications.
3. Let end-user to `exclude` them on at pleasure.

Any other ideas ?

sabbyanandan:
> So, from here it might be better to always provide the json-path dependency and allow picky users to exclude it in his environment.

I like the idea of including `jsonpath` and `spring-integration-xml` to ​_both_​ `<dependencies>` and `<dependencyManagement>` sections in the root pom.

Any OOTB or custom app that doesn't care for it can always exclude it easily anyway.

artembilan:
See the polishing for your confirmation: https://github.com/artembilan/spring-cloud-stream-modules/commit/e84b7e0e63e6f02c9ecbb74070b5cc2cdd3bbf6d

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-modules/commit/a17169f5542c054f6d0da03e5f322b7287f8841d.

Maybe we need such a story in the https://github.com/spring-cloud/spring-cloud-stream-app-starters?..

PR #253, opened 2016-04-15
Replace spring-cloud-stream-parent as main parent

### LINKED ISSUE #252: Do not set `spring-cloud-stream-parent` as parent
As a Spring Cloud Stream Modules developer, I do not want to depend on `spring-cloud-stream-parent` directly but to depend on its dependencies bom, so that I can have it manage Spring Cloud Stream dependendencies without inheriting its entire build lifecycle.

mbogoevici:
Fixes #252

Set `spring-cloud-build` as parent
Add `spring-cloud-stream-dependencies` as a dependencies BOM
Copy plugin configurations from the former parent, so that the current behaviour is preserved

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,8 +12,8 @@
 	</organization>
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
-		<artifactId>spring-cloud-stream-parent</artifactId>
-		<version>1.0.0.BUILD-SNAPSHOT</version>
+		<artifactId>spring-cloud-build</artifactId>
+		<version>1.1.0.RC1</version>
 		<relativePath />
 	</parent>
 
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
It looks like `spring-cloud-stream-modules-docs` still has `spring-cloud-stream-parent` as its parent and needs to be fixed as well.

ilayaperumalg:
Pushed this change: https://github.com/spring-cloud/spring-cloud-stream-modules/commit/1e0c349884b2944b2d04a17d472ccf337a9e3d90 that removes the `spring-cloud-stream-parent` reference for docs and add the `groupId` for the docs project without any parent configuration.

ilayaperumalg:
Also, pushed this change f49df9d7ec4d2fc28b5559f950cafe9466d39af0 to fix the artifiact Id for docs (wonder why this wasn't failing before).

PR #254, opened 2016-04-19
Add list operation for aggregate-counter

ilayaperumalg:


BoykoAlex:
Looks like `Redis` and `InMemory` Aggregate-Counter repos `#getCounts(...)` methods always return an AggregateCounter object, e.g. `InMemoryAggregateCounterRepository` always calls `getOrCreate(name)` method which creates counter if it doesn't exist.
I'm writing the get-and-delete unit test. Create counter `foo`, check the `get` with `/metrics/aggregate-counters/foo` then delete it with DELETE request. Deletion should check if counter is present first which is rather inconvinient now, beacuse only `list()` doesn't call `getOrCreate(name)`. Now, I need to check whether counter has been deleted, which is again inconvinient because `getCounts(...)` would always create a new counter if it's missing.
I concluded that`#exists(name)` on the aggregate counter repository would help I think :-)

ilayaperumalg:
@BoykoAlex I think the `find` REST call should ignore calling `getCounts` and the method `getCounts` should only be used to retrieve the aggreate-counter **values** instead of the aggregate counter itself.

for getAndDelete, you can always check if the `Set` that holds the aggregate counter names has the name in it and proceed to delete if it exists.

BoykoAlex:
There is an issue with `RedisAggregateCounterRepository#reset(String id)` method. If an aggregate-counter given it's id is deleted, the `list()` method and `getCounts` methods still return it :-(

ilayaperumalg:
@BoykoAlex interesting. The `reset` method call makes sure to delete the name from the `Set` which is used by the `list()` call. For the `getCounts` method, we should never call it as `finder` method as it does create the aggregate counter always.

sabbyanandan:
@ilayaperumalg: when you get a chance, please port this over to app-starters repo.

ilayaperumalg:
@sabbyanandan This was already ported there: https://github.com/spring-cloud/spring-cloud-stream-app-starters/pull/12

PR #255, opened 2016-04-19
Mail Source for Spring Cloud Data Flow

aamol:
Needed mail source module for my work as I am trying to migrate to Spring Cloud Data Flow from Spring XD. As the module was not yet ported from Spring XD created the module for our use. 

Raising pull request if code suits fine. Please feel free to provide suggestion/improvements

Thanks,
Amol

### DIFF
--- a/mail-source/README.adoc
+++ b/mail-source/README.adoc
@@ -1,0 +1,25 @@
+//tag::ref-doc[]
+= Http Source
+
+A source module that listens for HTTP requests and emits the body as a message payload.
+If the Content-Type matches `text/*` or `application/json`, the payload will be a String,
+otherwise the payload will be a byte array.
+
+== Options
+
+The **$$http$$** $$source$$ supports the following configuration properties:
+
+$$pathPattern$$:: $$An Ant-Style pattern to determine which http requests will be captured$$ *($$String$$, default: `/`)*
+
+//end::ref-doc[]
+== Build
+
+```
+$> mvn package
+```
+
+== Run
+
+```
+$> java -jar target/http-source-${version}-exec.jar
+```
--- a/mail-source/README.adoc
+++ b/mail-source/README.adoc
@@ -1,15 +1,22 @@
 //tag::ref-doc[]
-= Http Source
+= Mail Source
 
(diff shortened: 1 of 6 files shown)


artembilan:
See also https://jira.spring.io/browse/INT-3994.

@sabbyanandan , shouldn't we create new apps already in the `starters` project directly?

sabbyanandan:
@artembilan: That's right. 

@aamol: Thanks for the contribution! As an FYI: we are in the process of refactoring and moving all the Stream applications to [`spring-cloud-stream-app-starters`](https://github.com/spring-cloud/spring-cloud-stream-app-starters). `spring-cloud-stream-modules` repo would eventually become obsolete with the upcoming M3 release. 

That said, could you please rework this PR and submit to the app-starters repo instead? You could use the JMS source [commit](https://github.com/spring-cloud/spring-cloud-stream-app-starters/commit/dfbd5052a1a310f116ddf1b36c784431140d5c1b) and the test [commit](https://github.com/spring-cloud/spring-cloud-stream-app-starters/commit/abb5731183317d8b685671a5b163fa54bd34283b) as the reference.

aamol:
@sabbyanandan : Thanks, will work on the spring-cloud-stream-app-starters and submit PR there

@artembilan : I was the one who reported the issue and Gary provided a different way of handling it.Which worked fine.

artembilan:
@aamol ,

right, I know: https://github.com/spring-projects/spring-integration/pull/1771

And I am one of those who review that solution and already doesn't like, because of all those `DirectFieldAccessor` hell.

But I'm glad that you have found the solution with the `MailToStringTransformer` as good one.
Although that might not be suitable for all use-cases and we may consider to introduce more `AbstractMailMessageTransformer` implementations. E.g. `ExpressionMailMessageTransformer` or ever `GroovyMailMessageTransformer` to allow end-user to decide how to deal with that inconvenient mail API.

Anyway, I agree that for the first cut the `MailToStringTransformer` should be enough.

CC @garyrussell

sabbyanandan:
@aamol: Thanks for contributing this in the app-starters repo.

